PR #134, opened 2014-11-02
Replaced obsolete href="javascript:code" in favor of event listeners

Ginden:
Replaced obsolete href="javascript:code" in favor of modern event listeners.

### DIFF
--- a/index.html
+++ b/index.html
@@ -90,7 +90,7 @@
         Compiler</a>, which supports generators and many other ES6
         features through source transformation, but we would argue it
         compares favorably to
-        Traceur <a href="javascript:toggleComparison()">in several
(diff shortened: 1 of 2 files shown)


andyhu:
Could you provide more information on why `<a href="javascript:>" way is obsolete? I'm a bit curious on this, but can't find anything on the web. IMO `javascript:` is always better, since if you use hashes it will cause problems for some single page applications which they relies hashes in their router, it will change the url and also when you scrolled to a different position, it will reset your scroll status

facebook-github-bot-8:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

PR #150, opened 2014-12-07
Add try/catch helper to fix #148

### LINKED ISSUE #148: Possible try/catch optimizations in the runtime
I was reading through the runtime today and noticed the use of try/catch in [runtime.async][1] and [invoke][2] on two [places][3]. This will lead to deoptimizations and as they are core prices of functionality they are crucial for good performance. This can be fairly easily be avoided and I see two primary ways of doing the, either add a helper function like this:

```javascript
var _try = function _try(block, catcher) {
  try {
    block()
  } catch (err) {
    catcher(err)
  }
}
```

Or break out the mentioned sections of code into their own function to lessen the effect. I suggest the first, but it might be good to break out some of the code from `invoke` since it is currently ~110 lines long. What do you think about this? Which way do you think is best?

[1]: ../blob/master/runtime.js#L69-L74
[2]: ../blob/master/runtime.js#L104-L122
[3]: ../blob/master/runtime.js#L169-L202

maxnordlund:
This implements the first alternativ in #148, e.g. use a helper function for all try catch in the runtime.

### DIFF
--- a/runtime.js
+++ b/runtime.js
@@ -25,6 +25,16 @@
     return new Generator(innerFn, outerFn, self || null, tryList || []);
   }
   runtime.wrap = wrap;
+  
+  // Try/catch helper to minemize deoptimizations
+  var _try = function(block) {
+    try {
+      block();
+      return undefined;
+    } catch (err) {
+      return err;
+    }
+  };
 
   var GenStateSuspendedStart = "suspendedStart";
   var GenStateSuspendedYield = "suspendedYield";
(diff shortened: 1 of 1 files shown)


monsanto:
Any word on this? I've been wanting to take a performance pass over Regenerator and this is the first thing that came to mind.

maxnordlund:
Just rebased of master, should be cleanly mergeable.

benjamn:
Have any idea how much this helps performance?

maxnordlund:
I guess it depends, and without some sort of benchmark it's difficult to say. But this is an optimization that [bluebird][] uses, and is also mentioned on [html5rocks][].

[bluebird]: /petkaantonov/bluebird
[html5rocks]: http://www.html5rocks.com/en/tutorials/speed/v8/

PR #167, opened 2015-01-14
Allow the runtime to be required as a module, and make it strict-mode-compliant.

### LINKED ISSUE #165: Runtime is not strict mode compliant
To reproduce: `node --use_strict regenerator/runtime.js`. This is a problem when compiling with `6to5`, as it will automatically include the strict header.

###### Error Output
```
regenerator/runtime.js:318
        function next() {
        ^^^^^^^^
SyntaxError: In strict mode code, functions can only be declared at top level or immediately within another function.
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:929:3
```

###### Minor Performance Opportunity?
While on this topic, shouldn't the runtime include the `"use strict"` header so that it can take advantage of performance optimizations that v8 and other engines likely provide?

### LINKED ISSUE #166: Runtime defaults to creating a global variable
It's quite unfortunate that the runtime is not easy to package up as self-contained module. It seems to me that if the `exports` variable is available, then there should be no need to create the global variable.

I'm creating a self-contained library with generators, and it would be great to not have to manually delete this variable.

benjamn:
@sebmck @appden does this work for you?

I'm a little worried that this will be a breaking change for users who expect `regeneratorRuntime` to be defined globally for them. What if we kept defining it globally in https://github.com/facebook/regenerator/blob/master/runtime.js, but provided another file that you could `require` as a proper module instead?

Fixes #165.
Fixes #166.

### DIFF
--- a/runtime-module.js
+++ b/runtime-module.js
@@ -1,0 +1,20 @@
+// This method of obtaining a reference to the global object needs to be
+// kept identical to the way it is obtained in runtime.js
+var g =
+  typeof global === "object" ? global :
+  typeof window === "object" ? window : this;
+
+var hasOwn = Object.prototype.hasOwnProperty;
+var hadRuntime = hasOwn.call(g, "regeneratorRuntime");
+var oldRuntime = hadRuntime && g.regeneratorRuntime;
+delete g.regeneratorRuntime; // Force reevalutation of runtime.js.
+
+module.exports = require("./runtime");
+
+if (hadRuntime) {
+  // Restore the original runtime.
+  g.regeneratorRuntime = oldRuntime;
+} else {
+  // Remove the global property added by runtime.js.
+  delete g.regeneratorRuntime;
+}
--- a/runtime.js
+++ b/runtime.js
@@ -8,18 +8,30 @@
  * the same directory.
  */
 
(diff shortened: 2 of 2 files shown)


sebmck:
> What if we kept defining it globally in https://github.com/facebook/regenerator/blob/master/runtime.js, but provided another file that you could require as a proper module instead?

Sounds good to me.

appden:
Great, I'm excited by this change! I posted a comment on the commit that no longer appears to be there about the implicit global likely still violating strict mode.

runtime-module.js
```diff
@@ -0,0 +1,20 @@
+// This method of obtaining a reference to the global object needs to be
+// kept identical to the way it is obtained in runtime.js
+var g =
+  typeof global === "object" ? global :
+  typeof window === "window" ? window : this;
```
benjamn: Whoops, `"window"`??

fkling:
:+1:

Issue #173, opened 2015-02-10
return does not work properly inside try/finally


monsanto:
```
function *foo() {
    for(var i = 0; i < 2; i++) {
        try {
            return i
        } finally {}
    }
}

console.log(foo().next())
```

regenerator will run the loop to completion, and log `{value: 1, done: true}`. It should terminate the loop immediately and print `{value: 0, done: true}`, as io.js does.

benjamn:
This is definitely a bug! As you probably know, `finally` blocks have the ability to override abrupt control flow from the `try` block, but obviously that should only happen if the `finally` block causes an abrupt completion of its own.

I can look at this later today, but if you have time to submit a PR with a failing test case, that would make my work even easier.

monsanto:
Sure, I can cook up some test cases.

PR #174, opened 2015-02-10
add tests for try-finally + abrupt returns

monsanto:
This tests all manner of abrupt returns. 4 tests fail at the moment: honor return, honor break, honor continue, and override abrupt with continue. All tests added in this PR pass in iojs.

edit: I think fixing the first 3 will fix the 4th, it's just a consequence of how the tests are setup


monsanto:
Cross-ref #173

PR #177, opened 2015-02-20
Use the value for remote.path directly in fs.readFileSync

RichardLitt:
There is a bug in brfs which disallows passing variables to fs.readFileSync.
This means that anyone using this module with Browserify will have a breaking
build. This can be fixed by using the value of remote.path directly, which,
while not optimal, solves the issue while the bug for fs.readFileSync is fixed.

I'm not sure if this PR is the right way to go about it, or to wait until brfs
addresses the issue. This PR is what I am using to fix my current build.

See more here: https://github.com/substack/brfs/issues/36

### DIFF
--- a/main.js
+++ b/main.js
@@ -52,7 +52,7 @@ function compile(source, options) {
     return {
       // Shortcut: no generators or async functions to transform.
       code: (options.includeRuntime === true ? fs.readFileSync(
-        runtime.path, "utf-8"
+        path.join(__dirname, "runtime.js"), "utf-8"
       ) + "\n" : "") + source
     };
   }


benjamn:
If this helps with browserification, I see nothing wrong with it!

benjamn:
Published to NPM as v0.8.12: https://www.npmjs.com/package/regenerator

RichardLitt:
:+1: Awesome. Thanks!

RichardLitt:
Thanks @TooTallNate.

benjamn:
Thanks all, and sorry for the premature merge. I've published version v0.8.13 to NPM, which includes #178.

benjamn:
@sebmck @TooTallNate do you know of any way I can include tests for your projects in my Travis CI suite, to give better feedback on pull requests here?

benjamn:
Just added some tests that would have caught this.

RichardLitt:
:+1:

PR #178, opened 2015-02-20
Fix JS hoisting issue by using different var name

xaka:
This is the fix for bug introduced by https://github.com/facebook/regenerator/pull/177. Code reads file in sync way and it assumes `path` variable is pointing to `path` module, but as there is another `path` declaration within the same scope (`var path = new types.NodePath(ast);`), it gets hoisted on top and so `path` becomes `undefined`. It leads to following JS error:

```
TypeError: Cannot call method 'join' of undefined
```

### DIFF
--- a/main.js
+++ b/main.js
@@ -59,8 +59,8 @@ function compile(source, options) {
 
   var recastOptions = getRecastOptions(options);
   var ast = recast.parse(source, recastOptions);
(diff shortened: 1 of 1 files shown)


xaka:
Please re-publish package under different version as soon as this PR gets accepted. The builds are broken :)

Issue #195, opened 2015-05-01
delete g.regeneratorRuntime throws in IE


AutoSponge:
I've tried to track this down but I can't find the source to patch.

I suggest changing this:

```js
var hadRuntime = g.regeneratorRuntime && Object.getOwnPropertyNames(g).indexOf("regeneratorRuntime") >= 0;
var oldRuntime = hadRuntime && g.regeneratorRuntime;
delete g.regeneratorRuntime;
```

To this:

```js
var hadRuntime = g.regeneratorRuntime && Object.getOwnPropertyNames(g).indexOf("regeneratorRuntime") >= 0;
var oldRuntime = hadRuntime && g.regeneratorRuntime;
try {
  delete g.regeneratorRuntime;
} catch (e) {
  g.regeneratorRuntime = undefined;
}
```

It doesn't fail silently now in IE but this will.

sebmck:
Source is [runtime.js](https://github.com/facebook/regenerator/blob/master/runtime-module.js).

benjamn:
Fixed by #196.

PR #196, opened 2015-05-01
fixing IE error - issue #195

AutoSponge:
https://github.com/facebook/regenerator/issues/195


benjamn:
We only really need `g.regeneratorRuntime` to be falsy, so I simplified this a little: 227739c68f95924f7c192ab1178781d996b9139d

PR #197, opened 2015-05-04
Bind the step function to the active domain (if present)

mehcode:
This allows for libraries and applications 3 or 4 layers up to be able to use generators in a domain and have the domain context be persisted.

---

> In order to accept your pull request, we need you to submit a CLA. You only need to do this once, so if you've done this for another Facebook open source project, you're good to go. If you are submitting a pull request for the first time, just let us know that you have completed the CLA and we can cross-check with your GitHub username.

I've completed the CLA.

### DIFF
--- a/runtime.js
+++ b/runtime.js
@@ -121,7 +121,7 @@
         }
       };
 
-      if (process.domain) {
+      if (process && process.domain) {
         step = process.domain.bind(step);
       }
 
--- a/runtime.js
+++ b/runtime.js
@@ -121,7 +121,7 @@
         }
       };
 
-      if (process && process.domain) {
+      if (typeof process !== "undefined" && process.domain) {
         step = process.domain.bind(step);
       }
 
--- a/runtime.js
+++ b/runtime.js
@@ -105,10 +105,8 @@
   runtime.async = function(innerFn, outerFn, self, tryLocsList) {
     return new Promise(function(resolve, reject) {
       var generator = wrap(innerFn, outerFn, self, tryLocsList);
-      var callNext = step.bind(generator, "next");
-      var callThrow = step.bind(generator, "throw");
 
-      function step(method, arg) {
+      var step = function(method, arg) {
         var record = tryCatch(generator[method], generator, arg);
         if (record.type === "throw") {
           reject(record.arg);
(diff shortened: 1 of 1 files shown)


runtime.js
```diff
@@ -121,8 +119,15 @@
         } else {
           Promise.resolve(info.value).then(callNext, callThrow);
         }
+      };
+
+      if (process.domain) {
```
sebmck: `process` wont always be defined.

mehcode: Forgot about the browser environment; fixed

runtime.js
```diff
@@ -121,8 +119,15 @@
         } else {
           Promise.resolve(info.value).then(callNext, callThrow);
         }
+      };
+
+      if (process && process.domain) {
```
sebmck: This will still throw in browsers since `process` wont be defined. You'll need to do `typeof process !== "undefined"`.

fkling: This does not ensure that process is declared. It only ensures that process does not have a falsy value. It will still throw if it is not declared. typeof is pretty much the only way to go here.

fkling:
Not sure why our comments are marked as outdated. Either way, `process && process.domain` won't guard against undeclared `process`.

mehcode:
That makes sense. I should have known that. Fixed.

mehcode:
Is there something else I need to do to get this merged?

mehcode:
@benjamn ^ 

Not trying to be pushy but my team is stuck on this fork because our project uses domains and regenerator (through babel) and they'd like to get rid of the npm-shrinkwrap.json file.

benjamn:
Published to NPM as v0.8.27: http://npmjs.org/package/regenerator

benjamn:
At the time this pull request was merged, there were still lots of Node projects that hadn't stopped using `process.domain` (even though it was going to be deprecated). Relevant documentation and discussions from back then:

* https://nodejs.org/api/domain.html
* https://github.com/nodejs/node/issues/66
* https://github.com/nodejs/node/issues/10843
* https://nodejs.org/en/docs/guides/domain-postmortem/

It's probably safe to remove this now, if you want to submit a PR!

Andarist:
wow, ive commented directly on the commit, had no idea my comment would land in the issue thread, nice! :)

@benjamn thanks for the answer, I would love to send a PR, although I would include it as part of the https://github.com/facebook/regenerator/issues/302 if you decide to give it a green light, if not - im gonna send a separate, small PR to remove those lines

mehcode:
We're no longer using domains anymore, yeah. Don't have the time to send a PR but we won't notice it missing.

PR #207, opened 2015-06-25
Do not reset temp vars on Context init

kpdecker:
We know that they will not exist at this point so the overhead of iterating and running the has own property check is unnecessary.

This led to 4-16x speed increases for the Babel generator tests defined here:
http://kpdecker.github.io/six-speed/

### DIFF
--- a/runtime.js
+++ b/runtime.js
@@ -376,7 +376,7 @@
     // locations where there is no enclosing try statement.
     this.tryEntries = [{ tryLoc: "root" }];
     tryLocsList.forEach(pushTryEntry, this);
-    this.reset();
+    this.reset(true);
   }
 
   runtime.keys = function(object) {
@@ -449,7 +449,7 @@
   Context.prototype = {
     constructor: Context,
 
-    reset: function() {
+    reset: function(skipTempReset) {
       this.prev = 0;
       this.next = 0;
       this.sent = undefined;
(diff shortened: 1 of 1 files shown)


runtime.js
```diff
@@ -460,10 +460,12 @@
 
       // Pre-initialize at least 20 temporary variables to enable hidden
       // class optimizations for simple generators.
```
benjamn: Since performance was the only reason I originally bothered pre-initializing these fields, and you've demonstrated (with actual data) that it doesn't help, why don't we just replace this code with
```js
for (var tempIndex = 0, tempName;
     hasOwn.call(this, tempName = "t" + tempIndex);
     ++tempIndex) {
  this[tempName] = null;
}
```
That way, on the first run, there would be only one `hasOwn.call`, and no assignments. Does that make sense?

kpdecker: I had assumed that temp variable usage was not guaranteed to be in order. I.e. `t1` could exist and not `t0`, if this isn't the case then your version would be much faster than the original but potentially slightly slower than the flagged version.

benjamn: Ah yes, while it's true that temp variable names are generated in depth-first AST order, code execution is likely to be non-linear. This makes me think the original approach was always slightly broken if there were more than 20 temp variables…

How about something like this?
```js
for (var name in this) {
  // Not sure about the optimal order of these conditions:
  if (name.charAt(0) === "t" &&
      hasOwn.call(this, name) &&
      !isNaN(+name.slice(1)) {
    this[name] = undefined;
  }
}
```

kpdecker: Performance on that is quite bad when used without the flag, causing the performance relative to my "ES5 baseline" to drop from 1/4 to 1/26th. Perhaps that combined with the flag is the best option?

benjamn: Yeah, ok, let's keep the flag.

kpdecker: I've updated the PR per this discussion.

benjamn:
Thanks!

PR #208, opened 2015-06-25
Reset rather than delete sent field on next

kpdecker:
Delete was causing deoptimization due to backing class change and was unnecessary as this field was explicitly initialized to undefined and was not being checked for existence as best I can tell.

When combined with #207 saw a 40x throughput improvement under FF, from 39k operations/sec to 1.6M ops/src on http://kpdecker.github.io/six-speed/

### DIFF
--- a/runtime.js
+++ b/runtime.js
@@ -280,7 +280,7 @@
           if (state === GenStateSuspendedYield) {
             context.sent = arg;
           } else {
-            delete context.sent;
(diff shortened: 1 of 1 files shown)


kpdecker:
Similarly, under Chrome the execution went from 39k ops/sec to 674k ops/sec.

benjamn:
Thanks a ton! That speedup is impressive.

PR #220, opened 2015-10-01
Infinity loop problem

fanatid:
Regenerator has problem with *infinity loops*, for example, [sandbox](http://facebook.github.io/regenerator/) fail with next code:
```js
setTimeout(async function () {
  while (true) {
    await Promise.resolve()
  }
}, 0)
```

This PR solve this problem, also (probably) it's decreases memory consumption on sequence of await's.

### DIFF
--- a/runtime.js
+++ b/runtime.js
@@ -130,11 +130,12 @@
   }
 
   function AsyncIterator(generator) {
-    // This invoke function is written in a style that assumes some
-    // calling function (or Promise) will handle exceptions.
     function invoke(method, arg, resolve, reject) {
-      try {
-        var result = generator[method](arg);
+      var record = tryCatch(generator[method], generator, arg);
+      if (record.type === "throw") {
+        reject(record.arg);
+      } else {
+        var result = record.arg;
         var value = result.value;
         if (value instanceof AwaitArgument) {
           return Promise.resolve(value.arg).then(function(value) {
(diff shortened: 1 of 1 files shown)


benjamn:
What are the symptoms of this problem?

runtime.js
```diff
@@ -132,44 +132,53 @@
...
-            // into the generator, abandon iteration, whatever). With
-            // await, by contrast, there is no opportunity to examine the
-            // rejection reason outside the generator function, so the
-            // only option is to throw it from the await expression, and
-            // let the generator function handle the exception.
-            result.value = unwrapped;
-            return result;
+    function invoke(method, arg, resolve, reject) {
+      try {
+        var result = generator[method](arg);
```
benjamn: I believe this is the only line in this function that could throw, so I think you can use `tryCatch(generator, method, arg)` to avoid the explicit `try` block (which prevents some V8 optimizations).

fanatid:
Without this PR, *invoke* call create \*inifity\* unresolved `Promise` and after some time node fail with `Allocation failed`. But *invoke* can be represented like *tail call*, so we can implement *invoke* like *loop*.

example of \*infinity\* `Promise`:
```js
(function () {
  return Promise.resolve().then(arguments.callee)
})()
```

benjamn:
Thanks for spotting not this problem and providing a really clean solution!

benjamn:
Published [v0.8.41](https://github.com/facebook/regenerator/releases/tag/v0.8.41) to NPM.

PR #221, opened 2015-10-03
Support TypeCastExpression

splodingsocks:
Add a simple case statement to support parsing code with type cast expressions in them.

Should resolve: https://github.com/babel/babel/issues/2477

Just so that there's a record here, regenerator threw an error when trying to parse code with a type cast expression in it, like so: 

```javascript
function *test() {
  ((yield 4): number);
}
```

### DIFF
--- a/lib/emit.js
+++ b/lib/emit.js
@@ -1205,6 +1205,9 @@ Ep.explodeExpression = function(path, ignoreResult) {
 
     return self.contextProperty("sent");
 
+  case "TypeCastExpression":
+    return self.explodeExpression(path.get('expression'));
+
   default:
     throw new Error(
       "unknown Expression of type " +


splodingsocks:
On second thought, it looks like the real problem is that babel isn't stripping this flow annotation for some reason. I'll close this after I check that out some more.

benjamn:
When does the Flow transform run relative to Regenerator?

splodingsocks:
looks like Babel is running Regenerator first, based on some log statements I threw in the code. Now I'm trying to figure out where the order is specified.

splodingsocks:
Babel is running regenerator before flow is run. This should be more easily fixable once regenerator is a plugin, instead of hard-coded into babel:

https://github.com/babel/babel/pull/2483

PR #241, opened 2016-04-09
Minimise cases inside the regenerator FSM

shauns:
(NB: Also at babel/babel#3462 )

This PR aids downstream consumers of optimised code by producing a more idomatic switch statement in the regenerator FSM used for generators/async statements.

This greatly benefits code coverage tools that check for branch usage; as-is, it's not possible to cover every branch.

### DIFF
--- a/lib/emit.js
+++ b/lib/emit.js
@@ -234,6 +234,11 @@ Ep.getDispatchLoop = function() {
   // case, we can skip the rest of the statements until the next case.
   var alreadyEnded = false;
 
+  // Check if the listing entry is an abrupt return; if last thing in listing is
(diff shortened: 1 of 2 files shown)


shauns:
Found a use-case this doesn't cover--

benjamn:
Thanks for working on this! There's definitely room for lots of dead code elimination after the transform.

Two thoughts:
1. It would be good if this optimization could be turned off with an option, since debugging code that has had bits and pieces cleverly removed can be very frustrating.
2. Babel is currently using a fork of this code, so we either need to contribute the changes to that fork, or (my preference) un-fork the code so that facebook/regenerator remains the source of truth.

PR #242, opened 2016-04-11
add @@asyncIterator method to AsyncIterator.prototype

HyeonuPark:
based on [this section](https://tc39.github.io/proposal-async-iteration/#sec-asynciteratorprototype-asynciterator)


benjamn:
@HyeonuPark Can you please move these changes to https://github.com/facebook/regenerator/blob/master/packages/regenerator-runtime/runtime.js?

PR #252, opened 2016-06-15
Feature/use scoped promise

kkirby:
I had read that async functions are supposed to use the current scopes Promise constructor when creating Promises. I wasn't able to find this in the actual [proposal](https://tc39.github.io/ecmascript-asyncawait/) but I didn't extensively read it over. At any rate, I think it makes sense that an async function should use the current scope's Promise.

### DIFF
--- a/lib/visit.js
+++ b/lib/visit.js
@@ -138,13 +138,10 @@ var visitor = types.PathVisitor.fromMethodsObject({
       // outer function because they don't need it to be marked and don't
       // inherit from its .prototype.
       node.generator ? outerFnExpr : b.literal(null),
-      b.thisExpression()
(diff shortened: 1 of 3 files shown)


kittens:
> I had read that async functions are supposed to use the current scopes Promise constructor when creating Promises.

This isn't true.

> I wasn't able to find this in the actual proposal but I didn't extensively read it over.

The spec refers to the `%Promise%` intrinsic which cannot be overridden and is equivalent to the global `Promise`.

> At any rate, I think it makes sense that an async function should use the current scope's Promise.

This is a spec in-compliant change and wont happen.

kkirby:
@kittens Thanks for the clarification!

PR #222, opened 2015-10-05
Convert to a Babel plugin

kittens:
This is  WIP. There are currently ~8 failing tests so I thought I'd put it up for review before I look into them too much. This will heavily improve Babel performance and stability.  I know there are ways for regenerator to perform it's compilation in a single pass in the future too (right now there are a few visitors that are ran independently).

NOTE: This wont currently run unless you've got a `development` branch locally checked out and everything linked which is a bit of a PITA.

AFAIK this was :+1: between @benjamn and @amasad.

TODO:
- [ ] Add back `includeRuntime` option.
- [x] Add back `regenerator.transform` API.

PR #259, opened 2016-11-21
Reimplement using Babel plugins

benjamn:
Supersedes https://github.com/facebook/regenerator/pull/222. Thanks for getting that started, @kittens, and I'm glad it's finally (almost) done!

With this pull request, Regenerator now consists of three helper packages:

* [`regenerator-runtime`](https://github.com/facebook/regenerator/tree/use-babel-packages/packages/regenerator-runtime), which implements various runtime functions relied upon by generated code. Already [used by Babel](https://github.com/babel/babel/blob/7e020272c11a15fe86a1fc90087c6914a829cbd3/packages/babel-runtime/package.json#L10), with an upgrade [pending](https://github.com/babel/babel/pull/4877).
* [`regenerator-transform`](https://github.com/facebook/regenerator/tree/use-babel-packages/packages/regenerator-transform), a pure-Babel implementation of the Regenerator AST transform, suitable for independent use by any Babel compilation pipeline. Intended to replace and/or become the implementation of [`babel-plugin-transform-regenerator`](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-regenerator).
* [`regenerator-preset`](https://github.com/facebook/regenerator/tree/use-babel-packages/packages/regenerator-preset), a group of Babel plugins that enforce various preconditions of the Regenerator transform, such as transformation of block-scoped `let` and `const` declarations. See the full list [here](https://github.com/facebook/regenerator/blob/master/packages/regenerator-preset/index.js).

Why move the implementation of Regenerator back into this repository?

1. **Test coverage.** Believe it or not, [the Babel version](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-regenerator) doesn't run any of the numerous [tests](https://github.com/facebook/regenerator/blob/master/test/tests.es6.js) [from](https://github.com/facebook/regenerator/blob/master/test/async.es6.js) [this](https://github.com/facebook/regenerator/blob/master/test/tests.transform.js) [repository](https://github.com/facebook/regenerator/blob/use-babel-packages/test/sloppy.js), which inhibits safe forward development of Regenerator within the Babel repository.
2. **A place to file and fix bugs.** I would love to spend time investigating and fixing the [issues](https://github.com/facebook/regenerator/issues) that have been reported against this repository, and I believe this repository is a better place to track Regenerator bugs than https://github.com/babel/babel/issues, but that effort only makes sense for me if this repository represents the source code of Regenerator.
3. **Babel ecosystem compatibility.** Regenerator currently relies on various tools that have been deprecated and/or reimplemented by Babel, such as the [`defs`](https://www.npmjs.com/package/defs) compiler for block-scoped declarations, and my ad-hoc implementation of `for`-`of` loops. Reimplementing Regenerator as a Babel plugin allows it to take full advantage of Babel's highly-optimized plugin toolchain, without pulling in any unnecessary dependencies like [`recast`](https://www.npmjs.com/package/recast) and [`ast-types`](https://www.npmjs.com/package/ast-types).

In case you're not aware, the Babel implementation was originally adapted from this repository, which is perfectly fine according to the [`LICENSE`](https://github.com/facebook/regenerator/blob/master/LICENSE), but [the vast majority of meaningful commits](https://github.com/facebook/regenerator/graphs/contributors) in the history of this project still reside in this repository, so I believe this is where Regenerator ought to live.

cc @hzoo @loganfsmyth et al.


benjamn:
Since there are a bunch of commits here, [this commit](https://github.com/facebook/regenerator/pull/259/commits/905bd7c794935a881c722b98ff9c3b8d1dfd5a7e) is the real meat of this pull request.

benjamn:
Now that Regenerator is a multi-package project, I would love some guidance about using [Lerna](https://github.com/lerna/lerna) to wrangle the helper packages.

benjamn:
As there have been no objections, I've gone ahead and merged this. I will obviously bump the minor version before publishing!

PR #262, opened 2016-12-01
Actualize node versions

chicoxyzzy:
- Run CI only for LTS and stable versions
- drop unmaintained versions of Node.js from `package.json`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,7 @@
 language: node_js
 node_js:
+  - "stable"
   - "6.0"
-  - "5.0"
   - "4.0"
-  - "iojs"
-  - "0.12"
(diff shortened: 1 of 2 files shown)


benjamn:
Those older versions are a decent approximation of other engines with limited ECMAScript support, such as browers. Node 0.11, in particular, is the first Node version that introduced support for generator functions. So I'd like to keep running the tests with as many versions of Node as possible, and I've pushed a [commit](https://github.com/facebook/regenerator/commit/22996e61b9d51414879a3a029c4069965f871944) that should fix the tests that broke.

chicoxyzzy:
Anyway we should add Node 7 so I created #263

PR #265, opened 2016-12-08
compile with loose mode

hzoo:
If we move the implementation maybe we want to be as close as possible to what it was before?

https://github.com/babel/babel/blob/26b4e0909ef8ec3a45b08f2a475124ed83defdac/package.json#L51-L66

```
{
"comments": false,
    "presets": [
      [
        "es2015",
        {
          "loose": true
        }
      ],
      "stage-0"
    ],
    "plugins": [
      "./scripts/add-module-exports",
      "transform-runtime",
      "transform-class-properties",
      "transform-flow-strip-types"
    ],
}

the other plugins/presets won't matter if the syntax isn't used so we can ignore that.

### DIFF
--- a/packages/regenerator-transform/package.json
+++ b/packages/regenerator-transform/package.json
@@ -20,7 +20,7 @@
   },
   "babel": {
     "presets": [
-      ["es2015", { loose: true }]
+      ["es2015", { "loose": true }]
     ],
     "plugins": [
       "transform-runtime"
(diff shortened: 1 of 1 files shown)


hzoo:
thoughts on whether we also need to use add-module-exports or we do a require('x').default or we use imports syntax for https://github.com/babel/babel/pull/4881#discussion_r91558764?

PR #267, opened 2016-12-15
Fix for bug when transform-regenerator is run without transform-es2015-shorthand-properties

aickin:
I think I found a bug with `transformer-regenerator` when it is used without `transform-es2015-shorthand-properties`. If you take the following input code:

```
var o = {
  *foo() {
    return "foo";
  }
};
```
and apply **only** `transform-regenerator`, you get:
```
var o = {
  foo() {
        case 0:
          return _context.abrupt("return", "foo");

        case 1:
        case "end":
          return _context.stop();
      }
    }, _callee, this);
  }
};
```
Note that there is a reference to `_callee` at the end, but `_callee` is never defined. This causes a ReferenceError when the code is run.

If you add `transform-es2015-shorthand-properties` to the plugin list, you instead get this result:
```
var o = {
  foo: regeneratorRuntime.mark(function _callee() {
        case 0:
          return _context.abrupt("return", "foo");

        case 1:
        case "end":
          return _context.stop();
      }
    }, _callee, this);
  })
};
```
This version does define `_callee`, and it works.

I've been experimenting with doing different builds for different browsers, and there are some browser versions that support shorthand properties but not generators. It would be really nice to use `transform-regenerator` without `transform-es2015-shorthand-properties` in this case.

I made a patch for `transform-regenerator` in this PR that I believe fixes the problem. Since I don't really understand all the intricacies of `regenerator`, I decided to just translate all shorthand generator object methods into non-shorthand object properties with a generator function expression value before the rest of `transform-regenerator` runs. Since generator function expressions transform correctly already, this seems to work.

I added three test cases: 2 with shorthand object methods and 1 with a boring old object property. The shorthand object method test cases failed with the master branch and work with this PR. All other tests pass on my machine, but I'll watch CI to see how the patch does on other platforms.

I fully admit I don't understand all this code perfectly, and I welcome any and all feedback on the patch. Thanks for all your work!

### DIFF
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -12,56 +12,10 @@ import assert from "assert";
 import * as t from "babel-types";
 import { hoist } from "./hoist";
 import { Emitter } from "./emit";
+import replaceShorthandObjectMethod from "./replaceShorthandObjectMethod";
 import * as util from "./util";
 
 let getMarkInfo = require("private").makeAccessor();
-
-// we want to convert this shorthand object method into a normal object property
-// which is a function expression. for example, this:
-//  var foo = {
-//    *bar(baz) { return 5; }
-//  }
-// should be replaced with:
-//  var foo = {
-//    bar: function*(baz) { return 5; }
-//  }
-// to do this, we will clone the parameter array and the body of the function
-// into a new FunctionExpression.
-function replaceShorthandObjectMethod(path) {
-  if (!t.isObjectMethod(path.node)) {
-    return path;
-  }
-
-  // if it's not a generator method, we don't need to do anything.
-  if (!path.node.generator) {
-    return path;
-  }
-
-  const parameters = path.node.params.map(function (param) {
-    return t.cloneDeep(param);
-  })
-
-  if (!(path.node.key.name || path.node.key.value)) throw path.node.key;
-  const functionExpression = t.functionExpression(
-    null, // id
-    parameters, // params
-    t.cloneDeep(path.node.body), // body
-    path.node.generator,
-    path.node.async
-  );
-
-  path.replaceWith(
-    t.objectProperty(
-      t.cloneDeep(path.node.key), // key
-      functionExpression, //value
-      path.node.computed, // computed
-      false // shorthand
-    )
-  );
-
-  // return the function expression path.
-  return path.get("value");
-}
 
 exports.visitor = {
   Function: {
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -16,6 +16,53 @@ import * as util from "./util";
 
 let getMarkInfo = require("private").makeAccessor();
 
+// we want to convert this shorthand object method into a normal object property
+// which is a function expression. for example, this:
+//  var foo = {
+//    *bar(baz) { return 5; }
+//  }
+// should be replaced with:
+//  var foo = {
+//    bar: function*(baz) { return 5; }
+//  }
+// to do this, we will clone the parameter array and the body of the function
+// into a new FunctionExpression.
+function replaceShorthandObjectMethod(path) {
+  if (!t.isObjectMethod(path.node)) {
+    return path;
+  }
+
+  // if it's not a generator method, we don't need to do anything.
+  if (!path.node.generator) {
+    return path;
+  }
+
+  const parameters = path.node.params.map(function (param) {
+    return t.cloneDeep(param);
+  })
+
+  if (!(path.node.key.name || path.node.key.value)) throw path.node.key;
+  const functionExpression = t.functionExpression(
+    null, // id
+    parameters, // params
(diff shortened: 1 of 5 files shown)


aickin:
CI hasn't finished yet, but I thought to try the test suite on other versions of node. It fails on node 0.10 and node 0.12 because they don't understand some of the shorthand syntax used in my tests. I'd love some pointers on how best to fix that. Thanks!

aickin:
I spent a little more time on it and figured out a way to get tests to pass on all platforms. Essentially, I just made it so that the new shorthand property tests only run on node v4 and up, as node 0.12 and 0.10 don't support shorthand properties.

And it looks like CI passed, so I think this is ready for a review. Thanks!

test/tests-node4.es6.js
```diff
@@ -0,0 +1,38 @@
...
+
+describe("object properties", function () {
+  it("should work if the generator is a simple object property", function () {
+    var obj = {
+      gen: function*(x) {
+        yield x;
+      }
+    };
+
+    check(obj.gen("oyez"), ["oyez"]);
```
benjamn: Hah, you've picked up my word!

packages/regenerator-transform/src/visit.js
```diff
@@ -16,6 +16,53 @@ import * as util from "./util";
...
+    return path;
+  }
+
+  // if it's not a generator method, we don't need to do anything.
+  if (!path.node.generator) {
+    return path;
+  }
+
+  const parameters = path.node.params.map(function (param) {
+    return t.cloneDeep(param);
```
benjamn: How necessary are these deep clones (this one and the others)? Are we really worried about preserving the parts of the original AST that are being replaced?

packages/regenerator-transform/src/visit.js
```diff
@@ -16,6 +16,53 @@ import * as util from "./util";
...
+      t.cloneDeep(path.node.key), // key
+      functionExpression, //value
+      path.node.computed, // computed
+      false // shorthand
+    )
+  );
+
+  // return the function expression path.
+  return path.get("value");
+}
```
benjamn: Would you mind moving this logic into a separate module, then import `replaceShorthandObjectMethod`?

packages/regenerator-transform/src/visit.js
```diff
@@ -16,6 +16,53 @@ import * as util from "./util";
...
+  const functionExpression = t.functionExpression(
+    null, // id
+    parameters, // params
+    t.cloneDeep(path.node.body), // body
+    path.node.generator,
+    path.node.async
+  );
+
+  path.replaceWith(
+    t.objectProperty(
```
benjamn: Why create a new `ObjectProperty` instead of modifying the existing one? Does this have consequences for how Babel generates code?

test/tests-node4.es6.js
aickin: I try to match the project style. ;)

packages/regenerator-transform/src/visit.js
aickin: > Are we really worried about preserving the parts of the original AST that are being replaced?

I think we are very concerned about preserving the parts of the original function parameters and function body. It seems to me that any change to the function params or body could accidentally change the semantics of the code significantly.

> How necessary are these deep clones (this one and the others)? 

I chose to deepClone because I wasn't sure how babel responds if you use a single node object in two different ASTs (the original AST and the replacement AST). It felt like it might be OK, but I couldn't find any documentation that guaranteed that it would work, so cloning felt safer.

If you would like, I can try just reusing the parameter, body, and object key nodes in the new tree rather than cloning and see if the tests pass. Let me know.

packages/regenerator-transform/src/visit.js
aickin: There is no existing `ObjectProperty`, there's only an `ObjectMethod`. 

When you have something like:

```
{ foo: function() {} };
```

the AST looks like this (with some parts emitted, obviously):

```
ObjectExpression {
  properties: [
    ObjectMethod {
      name: "foo",
      params: [],
      body: BlockStatement {
        body:[]
      }
    }
  ]
}
```

The point of this function is to replace the `ObjectMethod` with an `ObjectProperty`.

For more info, here's an AST Explorer that shows how `ObjectMethod` and `ObjectProperty` show up in the AST: https://astexplorer.net/#/gist/f4ba155b3c70312a3d84d100a5294417/ab381e4c3cc009f2a88e6aca29f0cc972974b81d

packages/regenerator-transform/src/visit.js
aickin: I would be happy to!

aickin: Done, in 98d9a907ff86b0121c9b4e9dafc30c7b187f8885. I added some more comments and cleaned up the error handling a little as well.

aickin:
Thanks for the review! I believe I've responded to all the comments with either code revisions, explanations, or questions. Please let me know if you'd like me to change anything else!

Issue #274, opened 2017-01-31
yield * broken for Firefox < 45


AnilRedshift:
So this is a fun bug.
Try running this code through regenerator on FF 44 or below:
```
function *inner() { yield 1; }
function *outer() { yield *inner(); }

function runMe() {
    var g = outer();
    console.log(g.next());
    console.log(g.next());
    console.log(g.next());
}
```
You'll note that the generator never reports `done` as true. Here are the notes from my investigation. I am not 100% of the reason because some of the macro code in gecko is not easy to parse.

1. Firefox used to support some [alternative form](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features/The_legacy_Iterator_protocol) of the iterator.
2. It looks like they implemented the ES2015 version in FF 27, but then had some backcompat work to detect when code was trying to use the old protocol. 
3. Specifically, note that LegacyIteratorNext [always returns done: false](https://github.com/mozilla/gecko-dev/blob/0f254a30d684796bcc8b6e2a102a0095d25842bb/js/src/builtin/Iterator.js#L14)
4. Now note that any delegate relies on `next()` working in order to iterate through the delegate
```
  function values(iterable) {
    if (iterable) {
      var iteratorMethod = iterable[iteratorSymbol];
      if (iteratorMethod) {
        return iteratorMethod.call(iterable);
      }
```
where iteratorSymbol in this case is `Symbol.iterator`
5. The scene is almost set. `wrap` uses the `Generator.prototype` to create its functions:
```
  function wrap(innerFn, outerFn, self, tryLocsList) {
    // If outerFn provided and outerFn.prototype is a Generator, then outerFn.prototype instanceof Generator.
    var protoGenerator = outerFn && outerFn.prototype instanceof Generator ? outerFn : Generator;
    var generator = Object.create(protoGenerator.prototype);
```
6. And lastly (this is where some educated amount of speculation comes in), it looks like the iterator methods on the Generator prototype in FF < 45 [are](https://bugzilla.mozilla.org/show_bug.cgi?id=1201089) completely [busted](https://bugzilla.mozilla.org/show_bug.cgi?id=1225396).

So, putting it all together, this is what I believe happens:
Due to the bad Generator prototype, firefox decides that the code trying to get the iterator in `values()` is legacy and shoves it down the legacy iterator codepath. This means that `next` will ALWAYS return `done: false` unless a `StopIteration` exception is thrown.
[...]

PR #278, opened 2017-02-15
Resolve yield * issues for Firefox < 45

rayd:
Attempt to address #274. We're basically checking whether or not the `NativeIteratorPrototype`'s next function is the bad `LegacyIteratorNext` function. If it is, then we choose to use the polyfill. If it's not and we would have otherwise been using the native iterator, then we use the native iterator implementation.

Running the browser tests in `test/index.html` confirms that things work in Firefox < 45, in Firefox > 45 and in Chrome. Thoughts?

### DIFF
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -96,9 +96,13 @@
   if (NativeIteratorPrototype &&
       NativeIteratorPrototype !== Op &&
       hasOwn.call(NativeIteratorPrototype, iteratorSymbol)) {
-    // This environment has a native %IteratorPrototype%; use it instead
-    // of the polyfill.
-    IteratorPrototype = NativeIteratorPrototype;
+    // This environment has a native %IteratorPrototype%; use it instead of the
+    // polyfill unless it's the broken Firefox IteratorPrototype. See
+    // https://github.com/facebook/regenerator/issues/274 for more details.
+    var _n = NativeIteratorPrototype[iteratorSymbol].call(NativeIteratorPrototype);
+    if (!_n || !_n.next || _n.next.name !== 'LegacyIteratorNext') {
+      IteratorPrototype = NativeIteratorPrototype;
+    }
   }
 
   var Gp = GeneratorFunctionPrototype.prototype =
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -96,13 +96,9 @@
   if (NativeIteratorPrototype &&
       NativeIteratorPrototype !== Op &&
       hasOwn.call(NativeIteratorPrototype, iteratorSymbol)) {
-    // This environment has a native %IteratorPrototype%; use it instead of the
-    // polyfill unless it's the broken Firefox IteratorPrototype. See
-    // https://github.com/facebook/regenerator/issues/274 for more details.
-    var _n = NativeIteratorPrototype[iteratorSymbol].call(NativeIteratorPrototype);
-    if (!_n || !_n.next || _n.next.name !== 'LegacyIteratorNext') {
-      IteratorPrototype = NativeIteratorPrototype;
-    }
+    // This environment has a native %IteratorPrototype%; use it instead
+    // of the polyfill.
+    IteratorPrototype = NativeIteratorPrototype;
   }
 
   var Gp = GeneratorFunctionPrototype.prototype =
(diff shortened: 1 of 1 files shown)


AnilRedshift:
The approach seems reasonable to me. Another way of doing the detection instead of checking the name would be to make a simple iterator, call next() and see if it throws or returns done: true. If it throws, then it's the bad one.

I personally think either approach is reasonable, since this is a specific-legacy-browser workaround.

Thanks for providing a fix!

packages/regenerator-runtime/runtime.js
```diff
@@ -96,9 +96,13 @@
   if (NativeIteratorPrototype &&
       NativeIteratorPrototype !== Op &&
       hasOwn.call(NativeIteratorPrototype, iteratorSymbol)) {
-    // This environment has a native %IteratorPrototype%; use it instead
-    // of the polyfill.
-    IteratorPrototype = NativeIteratorPrototype;
+    // This environment has a native %IteratorPrototype%; use it instead of the
+    // polyfill unless it's the broken Firefox IteratorPrototype. See
+    // https://github.com/facebook/regenerator/issues/274 for more details.
+    var _n = NativeIteratorPrototype[iteratorSymbol].call(NativeIteratorPrototype);
```
benjamn: Can't this just be `NativeIteratorPrototype[iteratorSymbol]()`?

packages/regenerator-runtime/runtime.js
```diff
@@ -96,9 +96,13 @@
...
+    if (!_n || !_n.next || _n.next.name !== 'LegacyIteratorNext') {
```
benjamn: No need for obscure variable names (`_n`). Just call it something like `nativeIterator`, please.

benjamn:
Thanks for working on this! Please consider my requested changes, and also please evaluate @AnilRedshift's suggestion. I think I might prefer feature detection to function name testing, though I suppose this fix only needs to work in one browser.

rayd:
Good points - thanks for the feedback! Hopefully getting to this tonight!

rayd:
After a lot of digging, I think I understand what's happening now. Due to the changes introduced in #264, the prototype of `%GeneratorPrototype%` now matches the prototypes of the other iterator protoypes (i.e. the prototype of `Array Iterator`, `Map Iterator`, etc). In the problematic versions of FF, however, this prototype is the [legacy Iterator protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features/The_legacy_Iterator_protocol), which does not adhere to the standard iterator protocol. So this is why we're getting into the bad code path in FF < 45. It looks like after that version the prototype chain for iterators was updated to not include the legacy Iterator prototype. 

What's really causing the problems is that calling the `@@iterator` function on a generator returns the legacy `Iterator` rather than the generator object itself. This causes us to call the incorrect `next` function (the legacy iterator's, instead of the generator's). In newer versions of FF and in Chrome, that call to the `@@iterator` function instead returns the generator object itself, which means the iterator we're using adheres to the spec in those cases. And according to the spec, calling the `@@iterator` function should just return `this` ([see here(http://www.ecma-international.org/ecma-262/6.0/#sec-%iteratorprototype%-@@iterator)]), so it seems that calling the `@@iterator` function on a Generator should simply return the Generator object itself. There are a few solutions:

 1. Override the `@@iterator` function on the Generator prototype, so that it always returns the Generator object itself.
 2. Detect when we're in the above scenario and return the "right thing" from the `values()` function.
 3. Detect when we're in the above scenario and use the polyfill version of the iterator prototype, which shortcuts this issue.

Option 1 seems the most spec-compliant on all fronts (and is the simplest actually) and should resolve the issues, since the iterator object for "delegates" will be the Generator and not the legacy Iterator. And in my reading of the spec, I don't think defining `@@iterator` on the Generator prototype violates it.

AnilRedshift:
#1 seems like the right approach to me. Thanks for digging to the root cause

rayd:
Any thoughts on this @benjamn?

benjamn:
Just to be clear, by "the Generator prototype" do you mean the individual prototype that each generator function gets ([here](https://github.com/facebook/regenerator/blob/a176823333a51e1464ff30c89e6f2e6e1e7ac967/packages/regenerator-runtime/runtime.js#L140)), or the prototype shared by all generator objects (`Gp` in the runtime code)?

AnilRedshift:
Can we get this PR merged? :-)

benjamn:
Released! https://github.com/facebook/regenerator/releases/tag/runtime%400.10.5

Issue #255, opened 2016-08-08
Wrong return value


djelas:
The following code does not execute as expected:

``` js
function *example() {
  var value = 0;
  try {
    value = yield;
    value = yield;
  }
  finally {
    return value;
  }
}

var gen = example();
gen.next();
gen.next(1);

var returnValue = gen.return();
console.log("Returned: " + returnValue);
```

The returned value should be equal to 1, instead undefined is returned

benjamn:
Sorry for the delay. I've been waiting until #259 was merged to address pending issues on this repository.

Would you have time to submit a failing test case [here](https://github.com/facebook/regenerator/blob/master/test/tests.es6.js)?

PR #279, opened 2017-02-16
Added test for return value when overriden by finally

Jamesernator:
Basically this adds a test case for #255.

Currently fails for regenerator, works fine in latest Node/Chrome/Firefox.


benjamn:
Thanks, this is really helpful! I've merged it manually and added some additional similar (failing 😞) tests, which I'll push (along with a fix) today.

benjamn:
Manually merged: a9cd2ecba383db302d0fb33e40b80ebd3b34025f

Thanks again!

PR #281, opened 2017-02-23
Fix SPDX license expression.

goloroden:
In `package.json`, the `license` field was not clear, as it only mentioned `BSD`, but not whether you want BSD-2-Clause, -3-Clause, or -4-Clause. In the top-level `LICENSE` file it is obvious that you want BSD-2-Clause.

So, I've updated `package.json` accordingly with the correct SPDX license expression.

(You need this for automatically checking license compatibility, so it would be great if you could merge this 😊)


Andarist:
@benjamn ping ;)

danez:
This can be closed now as of #334

PR #283, opened 2017-03-16
Fix hoisting of arguments

danez:
In the case that the generator or the async function is initially an arrow function and uses only a rest param, the babel es2015 transforms will replace the arrow with a function and the rest param with arguments. regenerator-transform will later also replace and move the arguments. In the end babel will hoist the arguments back up, but to the wrong level.

Thats a little bit complicated to describe, maybe better I show it with code:

1. initial
```js
function test(fn) {
  return async (...args) => {
    return fn(...args);
  };
}
```

2. arrow transform
```js
function test(fn) {
  return async function (...args) {
    return fn(...args);
  };
}
```

3. rest transform
```js
function test(fn) {
  return async function () {
    return fn.apply(this, arguments);
  };
}
```

4. regenerator transform (this is what it should be in the end)
```js
function test(fn) {
  var _this = this;
  return function () {
    var _ref = _asyncToGenerator(regeneratorRuntime.mark(function _callee() {
      var _args = arguments;
        while (1) {
            case 0:
              return _context.abrupt("return", fn.apply(undefined, _args));

            case 1:
            case "end":
              return _context.stop();
          }
        }
      }, _callee, _this);
    }));

    return function () {
      return _ref.apply(this, arguments);
    };
  }();
}
```

5. babel hoist (not exactly sure what is triggering this, maybe another plugin, or babel itself)

> Note that arguments gets hoisted incorrectly

```js
function test(fn) {
  var _arguments = arguments,
      _this = this;

  return function () {
    var _ref = _asyncToGenerator(regeneratorRuntime.mark(function _callee() {
      var _args = _arguments;
        while (1) {
            case 0:
              return _context.abrupt("return", fn.apply(undefined, _args));

            case 1:
            case "end":
              return _context.stop();
          }
        }
      }, _callee, _this);
    }));

    return function () {
      return _ref.apply(this, arguments);
    };
  }();
}
```

This change sets a hint on the `arguments`-identifier to which function it belongs to, and it will only get hoisted up to this level.
We also do this in babel in certain places: https://github.com/babel/babel/blob/7.0/packages/babel-plugin-transform-es2015-parameters/src/rest.js#L212

I tried creating an test, but couldn't find exactly how to do it.

//cc @hzoo @existentialism @loganfsmyth 

Fixes: babel/babel#5485 babel/babel#5330 babel/babel#4219

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,8 @@
     "through": "^2.3.8"
   },
   "devDependencies": {
+    "babel-plugin-transform-es2015-parameters": "^6.23.0",
+    "babel-plugin-transform-es2015-spread": "^6.22.0",
     "babylon": "^6.14.1",
     "mocha": "^2.3.4",
     "promise": "^7.0.4",
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -96,8 +96,11 @@ exports.visitor = {
       let didRenameArguments = renameArguments(path, argsId);
       if (didRenameArguments) {
         vars = vars || t.variableDeclaration("var", []);
+        const argumentIdentifier = t.identifier("arguments");
+        // we need to do this as otherwise arguments in arrow functions gets hoisted
+        argumentIdentifier._shadowedFunctionLiteral = path;
         vars.declarations.push(t.variableDeclarator(
-          argsId, t.identifier("arguments")
+          argsId, argumentIdentifier
         ));
       }
 
--- a/test/regression.js
+++ b/test/regression.js
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) 2017, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * https://raw.github.com/facebook/regenerator/master/LICENSE file. An
+ * additional grant of patent rights can be found in the PATENTS file in
+ * the same directory.
+ */
+
+var assert = require("assert");
+
+describe("regressions", function() {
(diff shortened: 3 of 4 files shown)


benjamn:
I'm not an expert on how hoisting works in Babel, though I have encountered it before. If anyone else has immediate ideas, I'm curious too!

Kovensky:
Is this possibly related to https://github.com/gaearon/react-hot-loader/issues/391 ?

danez:
I was thinking the same that there is probably the same issue with this. But this PR is only for arguments. Maybe I can figure out if we can do the same with this.

@benjamn yeah it is weird that we have to use a private flag, but not sure what else we should do in this case, maybe we can disable the hoisting in a better way, I will investigate a little more.

danez:
@Kovensky The `this` issues  in 5078 seems unrelated, and is just an issue with copying the the nodes. 

I tried different variants with `this`and regenerator and they all work. 

Another idea I had was turning the function into a non shadowed function by doing `delete node.shadow`, but this seems more risky to break other stuff.

benjamn:
@danez Care to add a test for this behavior, just in case the `_shadowedFunctionLiteral` "API" ever changes without warning?

danez:
If you tell me how. All the tests seem to require native support for the features being tested, which might be hard here for `async` functions.

benjamn:
Sure, the `test/async.js` tests should run only in translation: https://github.com/facebook/regenerator/blob/master/test/async.js

danez:
Thanks for the help, hope it is okay they way I setup the test.

PR #284, opened 2017-03-18
Use babel-preset-env instead of babel-preset-es2015

danez:
This includes support for es2016 & es2017 and we are also more or less recommending everyone to use env instead of `latest`- or `es20XX`-presets.

If at some point node <4 might be dropped in this package, we could also easily target node4 as environment, which excludes some transforms.

Right now the transformed sourcecode is exactly the same.


benjamn:
Awesome :tada:

PR #290, opened 2017-04-26
Make AsyncIterator comply with AsyncIterable

leebyron:
Similar to what is defined by Iterator Prototype.

See: https://tc39.github.io/proposal-async-iteration/#sec-asynciteratorprototype-asynciterator


benjamn:
Thanks @leebyron!

Issue #112, opened 2014-04-30
Make runtime work within CSP restriction.


creationix:
In particular, I want to use regenerator to write chrome packaged apps like my tedit editor.  But as soon as I tried this I realized that the runtime uses eval somewhere.

Is this a hard requirement of the runtime, or can it be modified to work without using eval (or it's variants like new Function, etc)?

I would love to use this for authoring chrome apps, but the last road-block is the runtime triggering CSP violations

PR #292, opened 2017-04-28
Simplify reference to the global object

maxnordlund:
Take advantage of sloppy mode, aka non-strict mode, where unbound functions are called with the global object as `this`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Securing_JavaScript

### DIFF
--- a/packages/regenerator-runtime/runtime-module.js
+++ b/packages/regenerator-runtime/runtime-module.js
@@ -1,9 +1,6 @@
 // This method of obtaining a reference to the global object needs to be
 // kept identical to the way it is obtained in runtime.js
-var g =
-  typeof global === "object" ? global :
-  typeof window === "object" ? window :
-  typeof self === "object" ? self : this;
+var g = (function() { return this })() || Function("return this")();
 
 // Use `getOwnPropertyNames` because not all browsers support calling
 // `hasOwnProperty` on the global `self` object in a worker. See #183.
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -727,10 +727,8 @@
     }
   };
 })(
-  // Among the various tricks for obtaining a reference to the global
-  // object, this seems to be the most reliable technique that does not
-  // use indirect eval (which violates Content Security Policy).
-  typeof global === "object" ? global :
-  typeof window === "object" ? window :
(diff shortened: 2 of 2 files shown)


benjamn:
See also: https://github.com/facebook/regenerator/issues/112

It's good that this doesn't use `Function` (for CSP reasons), but It seems like this could fail if the runtime is embedded in a bundle of strict code?

benjamn:
What about this?
```js
(function { return this }()) || Function("return this")()
```
This avoids using the `Function` constructor unless absolutely necessary, but still allows the developer to avoid the CSP violation of using the `Function` constructor.

packages/regenerator-runtime/runtime-module.js
```diff
@@ -1,9 +1,6 @@
 // This method of obtaining a reference to the global object needs to be
 // kept identical to the way it is obtained in runtime.js
-var g =
-  typeof global === "object" ? global :
-  typeof window === "object" ? window :
-  typeof self === "object" ? self : this;
+var g = (function { return this })();
```
benjamn: An anonymous and parameter-less function?? Think you need a `()`, as the tests will surely tell you.

benjamn: Wow, I guess none of the tests use `runtime-module.js`, huh.

maxnordlund: Darn it, knew I forgot something. I'll fix it right away.

maxnordlund:
Yeah, the strict code thing is real. But it should be safe as long as you don't blindly concatenate all your JS together with the `"use strict";` at the top, as [amazon](https://bugzilla.mozilla.org/show_bug.cgi?id=579119) did by mistake.

I think most, if not all, users of this library uses a bundler, which I hope uses a function closure to keep things separate. Then it's fine if something else is using top level strict mode.

> What about this?
```javascript
(function(){return this}()) || Function("return this")()
```

That looks like the ticket. I'll get it in, and fix the other thing you pointed out.

maxnordlund:
I'm implemented that fix, and squashed the commits while I was at it.

benjamn:
Thanks for working on this, and sorry it sat unmerged for so long!

maxnordlund:
No worries, life happens.

PR #293, opened 2017-04-28
Don't use duck typing for awaitable test

maxnordlund:
Since the user might yield a value with a __await key, it's impossible to distinguish from the intended use case.

Instead use a private class and the `instanceof` operator, which is much more reliable.

_Though the new [`Symbol.hasInstance`][1] certainly makes things more interesting._

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance

### DIFF
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -144,10 +144,14 @@
 
   // Within the body of any async function, `await x` is transformed to
   // `yield regeneratorRuntime.awrap(x)`, so that the runtime can test
-  // `hasOwn.call(value, "__await")` to determine if the yielded value is
+  // `value instanceof __Awaitable` to determine if the yielded value is
   // meant to be awaited.
   runtime.awrap = function(arg) {
-    return { __await: arg };
+    return new __Awaitable(arg)
+  };
+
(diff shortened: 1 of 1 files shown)


benjamn:
What you're proposing is in fact how `runtime.awrap` used to work, and the current implementation was a deliberate change to avoid the many problems with `instanceof`.

If you dig into the blame history, you'll find this relevant commit: https://github.com/maxnordlund/regenerator/commit/e62a7e500885211aa5bce1b6110d9a77aecc0c34

This is also the convention used by `async-to-gen`, and I think it's important for code generated by either library to work with the other: https://github.com/leebyron/async-to-gen/blob/88ffec6e181e8add71a012c8fca2c5b0f4d0c9eb/index.js#L508

> Since the user might yield a value with a __await key, it's impossible to distinguish from the intended use case.

That's really not something I'm worried about, especially since this convention is used outside Regenerator. The possibility of intentionally yielding `{ __await: ... }` (without having access to a private constructor) is specifically what allows different code generators to work together.

maxnordlund:
I see, well then there are good reasons for using duck typing. I'll just close this then.

PR #294, opened 2017-05-06
Don't pass falsy nodes to `path.replaceWith()`

josephfrazier:
Before, when regenerator is run with the following input:

```js
async function foo() {
  for (let i; i;);
}
```

it throws this error:

    You passed `path.replaceWith()` a falsy node, use `path.remove()` instead

With this change, the following code is output:

```js
function foo() {
  var i;
  return regeneratorRuntime.async(function foo$(_context) {
    while (1) {
        case 0:
          for (; i;) {}

        case 1:
        case "end":
          return _context.stop();
      }
    }
  }, null, this);
}
```

Fixes https://github.com/babel/babel/issues/4953

### DIFF
--- a/packages/regenerator-transform/src/emit.js
+++ b/packages/regenerator-transform/src/emit.js
@@ -601,7 +601,7 @@ Ep.explodeStatement = function(path, labelId) {
     }
 
     let discriminant = path.get("discriminant");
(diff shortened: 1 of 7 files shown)


benjamn:
Thanks @josephfrazier!

PR #300, opened 2017-07-03
fix #299, check that Symbol is not a polyfill

### LINKED ISSUE #299: Broken with polyfill.io
regenerator-runtime: 0.10.5
chrome version 44.0.2403

[This code](https://github.com/facebook/regenerator/blob/ea4de31e4a8151fab67dfa7485b2ac5ede4f8eb3/packages/regenerator-runtime/runtime.js#L417):

```
Gp[toStringTagSymbol] = "Generator";
```

throws:

`Uncaught TypeError: Cannot assign to read only property '__symbol:toStringTag0.123....' of [object Object]`

It doesn't happen in latest chrome.
It's Android users, so we can't ask them to update their browsers...

Do you have any lead that could resolve this ?

Thank you

christophehurpeau:
fix #299

### DIFF
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -14,7 +14,7 @@
   var Op = Object.prototype;
   var hasOwn = Op.hasOwnProperty;
   var undefined; // More compressible than void 0.
-  var $Symbol = typeof Symbol === "function" ? Symbol : {};
+  var $Symbol = typeof Symbol === "function" && typeof Symbol.toStringTag === "symbol" ? Symbol : {};
   var iteratorSymbol = $Symbol.iterator || "@@iterator";
   var asyncIteratorSymbol = $Symbol.asyncIterator || "@@asyncIterator";
   var toStringTagSymbol = $Symbol.toStringTag || "@@toStringTag";


fstoerkle:
Is there a reason why this PR was closed? We are currently facing the same issue and what love to get that resolved.

christophehurpeau:
for some reason this didn't work, I "fixed" it by doing:

```
<script>window.__SYMBOL_POLYFILL = !window.Symbol || !!window.Symbol.toStringTag</script>
<script src="https://polyfill.io/v2/polyfill.min.js?features=..."></script>
```

And then require regenerator-runtime:

```
if (!global.__SYMBOL_POLYFILL) {
  const globalPolyfillSymbol = Symbol;
  global.Symbol = undefined;
  require('regenerator-runtime/runtime');
  global.Symbol = globalPolyfillSymbol;
} else {
  require('regenerator-runtime/runtime');
}
```

fstoerkle:
Thanks for the update @christophehurpeau. We might work on this later this/next week, let's see if we come up with something

christophehurpeau:
I found out why it didn't work before: I simply had 2 versions of regenerator-runtime in my node_modules.

christophehurpeau:
The build fails beacause `Cannot find module 'regenerator-transform'`, like other recent PR

https://travis-ci.org/facebook/regenerator/pull_requests

zloirock:
It will break compatibility with `core-js`.

christophehurpeau:
fix in polyfill-service : https://github.com/Financial-Times/polyfill-service/pull/1295

Issue #286, opened 2017-04-10
Tree-shaking transformed generator functions


rjdestigter:
```
export function *xyz() {}

export function *abc() {}
```

compiles to:
```"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.test = test;
exports.abc = abc;

var _marked = [test, abc].map(regeneratorRuntime.mark);

function test() {
  return regeneratorRuntime.wrap(function test$(_context) {
    while (1) {
      switch (_context.prev = _context.next) {
        case 0:
        case "end":
          return _context.stop();
      }
    }
  }, _marked[0], this);
}

function abc() {
  return regeneratorRuntime.wrap(function abc$(_context2) {
    while (1) {
      switch (_context2.prev = _context2.next) {
        case 0:
        case "end":
          return _context2.stop();
      }
    }
  }, _marked[1], this);
}
```

Because of the line `var _marked = [test, abc].map(regeneratorRuntime.mark);` using `Webpack 2` + `UglifyJS` does not _tree shake_  if say only one of the functions is imported in the application.

I wouldn't mind submitting a PR request if someone could point me in the right direction (file) and we think changing this would be a nice update.

benjamn:
This is tricky because `regenerator-transform` runs long before any tree-shaking takes place, so I'm not sure there's anything Regenerator can do to help with this, by itself. Any ideas?

rjdestigter:
What if it compiled to:
```
var _marked_test = regeneratorRuntime.mark(test);
var _marked_abc = regeneratorRuntime.mark(abc);

function abc() {
  return regeneratorRuntime.wrap(function abc$...
    ...
  }, _marked_abc, this);
}

function test() {
  return regeneratorRuntime.wrap(function test$...
    ...
  }, _marked_test, this);
}
```

benjamn:
That would definitely be equivalent, and it would probably be worth the repetition if it helped with tree shaking. Is Webpack smart enough to remove the whole `var _marked_abc = regeneratorRuntime.mark(abc);` line if `_marked_abc` is no longer used?

rjdestigter:
Yes, I'm double checking/testing it at the moment but I'm pretty sure it does

benjamn:
I'm a little worried that `regeneratorRuntime.mark(abc)` by itself is going to prevent the `abc` function from being removed in the first place, but I'd be happy to be wrong!

rjdestigter:
I put the following code inside one module: `generators.js`

[...]

PR #301, opened 2017-07-21
refactor mark declaration to make it more tree-shakable (#286)

modosc:
per the conversation on #286 this should make my bundles much smaller. 

i haven't tested in our application yet, getting all the deps installed right is difficult (even getting the tests running right pretty unclear). i plan on doing so tomorrow to verify but it sounds like others have done this already?

### DIFF
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -86,10 +86,10 @@ exports.visitor = {
 
       // it's possible that getOuterFnExprPath removes our current path so we
       // reassign with the return value
-      path = getOuterFnExprPath(path)
+      path = getOuterFnExprPath(path);
       // and also reassign node as well
-      node = path.node
-      let outerFnExpr = node.id
+      node = path.node;
+      let outerFnExpr = node.id;
 
       // Note that getOuterFnExprPath has the side-effect of ensuring that the
       // function has a name (so node.id will always be an Identifier), even
@@ -187,7 +187,7 @@ function getOuterFnExprPath(funPath) {
   let outerFnExpr = node.id
 
   if (node.generator && // Non-generator functions don't need to be marked.
-    t.isFunctionDeclaration(node)) {
+      t.isFunctionDeclaration(node)) {
     let pp = funPath.findParent(function(path) {
       return path.isProgram() || path.isBlockStatement();
     });
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -84,8 +84,14 @@ exports.visitor = {
         bodyBlockPath.node.body = innerBody;
       }
 
-      let outerFnExpr = getOuterFnExpr(path);
-      // Note that getOuterFnExpr has the side-effect of ensuring that the
+      // it's possible that getOuterFnExprPath removes our current path so we
+      // reassign with the return value
+      path = getOuterFnExprPath(path)
+      // and also reassign node as well
+      node = path.node
+      let outerFnExpr = node.id
+
+      // Note that getOuterFnExprPath has the side-effect of ensuring that the
       // function has a name (so node.id will always be an Identifier), even
       // if a temporary name has to be synthesized.
       t.assertIdentifier(node.id);
@@ -152,7 +158,8 @@ exports.visitor = {
       }
 
       if (wasGeneratorFunction && t.isExpression(node)) {
-        util.replaceWithOrRemove(path, t.callExpression(util.runtimeProperty("mark"), [node]))
+        util.replaceWithOrRemove(path, t.callExpression(util.runtimeProperty("mark"), [node]));
+        path.addComment("leading", "#__PURE__");
       }
 
       // Generators are processed in 'exit' handlers so that regenerator only has to run on
@@ -163,11 +170,11 @@ exports.visitor = {
   }
 };
 
-// Given a NodePath for a Function, return an Expression node that can be
+// Given a NodePath for a Function, return a path to an Expression node that can be
 // used to refer reliably to the function object from inside the function.
 // This expression is essentially a replacement for arguments.callee, with
 // the key advantage that it works in strict mode.
-function getOuterFnExpr(funPath) {
+function getOuterFnExprPath(funPath) {
   let node = funPath.node;
   t.assertFunction(node);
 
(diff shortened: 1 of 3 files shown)


modosc:
it looks like there's a change in [npm 5's link behavior](http://codetunnel.io/npm-5-changes-to-npm-link/) that's causing the build to [break on node 8](https://travis-ci.org/facebook/regenerator/jobs/255876682). i saw this behavior locally too, its why i had such a hard time testing. i'll look into it tomorrow.

benjamn:
@modosc Thanks for working on this!

modosc:
this works now.

modosc:
ping?

test/tests.transform.js
```diff
@@ -3,7 +3,50 @@ var recast = require("recast");
...
+describe("pure", function() {
+  it("should work with a function expression", function() {
+    var ast = recast.parse('var a = function* foo(){};', {
+        parser: require("babylon")
+    });
+    const code = recast.print(transform(ast)).code;
+    assert.ok(
+      code.indexOf(
+        'var a = /*#__PURE__*/regeneratorRuntime.mark(function foo() {'
+      ) !== 1
```
benjamn: I would use `assert.notStrictEqual` here (style) and compare against `-1` rather than `1` (correctness).

packages/regenerator-transform/src/visit.js
```diff
@@ -84,8 +84,14 @@ exports.visitor = {
...
       }
 
-      let outerFnExpr = getOuterFnExpr(path);
-      // Note that getOuterFnExpr has the side-effect of ensuring that the
+      // it's possible that getOuterFnExprPath removes our current path so we
+      // reassign with the return value
+      path = getOuterFnExprPath(path)
+      // and also reassign node as well
+      node = path.node
+      let outerFnExpr = node.id
```
benjamn: The code style for this repository is to use semicolons, so please do (sorry)!

packages/regenerator-transform/src/visit.js
```diff
@@ -177,32 +184,118 @@ function getOuterFnExpr(funPath) {
     node.id = funPath.scope.parent.generateUidIdentifier("callee");
   }
 
+  let outerFnExpr = node.id
+
   if (node.generator && // Non-generator functions don't need to be marked.
-      t.isFunctionDeclaration(node)) {
-    let pp = funPath.findParent(function (path) {
+    t.isFunctionDeclaration(node)) {
```
benjamn: Please line up the conditional expressions vertically, like before.

packages/regenerator-transform/src/visit.js
```diff
@@ -177,32 +184,118 @@ function getOuterFnExpr(funPath) {
...
 
-    let markDecl = getRuntimeMarkDecl(pp);
-    let markedArray = markDecl.declarations[0].id;
-    let funDeclIdArray = markDecl.declarations[0].init.callee.object;
-    t.assertArrayExpression(funDeclIdArray);
+    // in spite of the above, we want the resulting function's name property
+    // property to be the same as the original so we wrap the FunctionExpression
+    // with:
+    //   Object.defineProperty(functionExpression, 'name', { value: outerFnExpr.name }
+    // this way introspection / prototype chains work the way we'd expect
```
benjamn: If I recall correctly, setting the `.name` property of functions doesn't work in all browsers. Can we split the `Object.defineProperty` stuff into a separate PR, or is that important for tree shaking?

benjamn: Now that I think about it, we're generating the code here, so there should be no need to call `Object.defineProperty` at runtime to set the `.name`—why not just generate a function with the right name?

packages/regenerator-transform/src/visit.js
```diff
@@ -177,32 +184,118 @@ function getOuterFnExpr(funPath) {
...
+      return t.isVariableDeclaration(p) && p.declarations === declarations;
+    })
+    assert(!isNaN(declarationsIndex));
+
+    // now push our new declaration into our list
+    declarations.push(decl)
+
+    // special case - handle node types that take FunctionDeclarations as
+    // arguments. not sure if there are more than these two, i couldn't find any
+    // in the docs?
```
benjamn: I'm hoping these cases become unnecessary if we simplify or remove the `name` logic above. 🤞

packages/regenerator-transform/src/visit.js
modosc: i don't think that would work because then `funtion abc()` will shadow `var abc` in the call to `regeneratorRuntime.wrap(`:
```
var abc = regeneratorRuntime.mark(function abc() { 
  return regeneratorRuntime.wrap(function abc$() { 
  }, abc, this); 
});
```

modosc: also afaict using `Object.defineProperty` shouldn't be an issue:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Browser_compatibility

test/tests.transform.js
modosc: done

packages/regenerator-transform/src/visit.js
benjamn: The `regeneratorRuntime.mark` function just returns the function you pass to it, so the two `abc` bindings should refer to the same function object, so the shadowing shouldn't be a problem. Even if the outer value of `abc` gets reassigned, I don't think it's important for `regeneratorRuntime.wrap` to get the new value.

The problem isn't `Object.defineProperty` support (though older IEs mess that up, too), but setting the `name` property on a function object, which either silently fails or throws an exception in some older browsers. It's also just a shame to do something at runtime that could be done at compile time.

packages/regenerator-transform/src/visit.js
modosc: done.

packages/regenerator-transform/src/visit.js
modosc: done.

benjamn:
Would you mind explaining in the PR description (above) exactly what's different about the generated code, with examples? I'm having trouble keeping track of the various parts of the new design. Thanks!

modosc:
one other issue - i realized i mean to skip the pure tests - i can't seem to get the tests to generate code with comments in it. 

the same code works correctly via `bin/regenerator` so i figured there was some option i was missing?

packages/regenerator-transform/src/visit.js
modosc: ok - let me try it out and verify before i update the description. 

thanks for taking the time to review btw.

test/tests.transform.js
Andarist: Imho while this type of test is easier to write quickly we should focus more on testing the actual outcome of the change - that those can be tree shaken properly. Imho example code should get bundled + minified and the bundled code should be checked against not having those in it.

modosc: sure - i just wanted to verify that the comment is being added in the write place. i noted below that i'm not able to do this now, comments disappear from the test generated code.

packages/regenerator-transform/src/visit.js
modosc: i made the change locally - all tests pass except [this one](https://github.com/facebook/regenerator/blob/master/test/tests.es6.js#L1579):
```
  1) function declaration hoisting should not interfere with function rebinding:
     AssertionError [ERR_ASSERTION]: [GeneratorFunction: toBeRebound] deepEqual 42
      at check (test/shared.js:16:12)
      at Context.<anonymous> (test/tests.es5.js:3307:5)
```

i _think_ it's because of the shadowing issue i mentioned. the original transform generates this code:
```
var _marked = [toBeRebound].map(regeneratorRuntime.mark);

function rebindTo(value) {
  var oldValue = toBeRebound;
  toBeRebound = value;
  return oldValue;
}

function toBeRebound() {
  var originalValue;
  return regeneratorRuntime.wrap(function toBeRebound$(_context) {
    while (1) {
        case 0:
          originalValue = toBeRebound;
          _context.next = 3;
          return toBeRebound;

        case 3:
          assert.strictEqual(rebindTo(42), originalValue);
          _context.next = 6;
          return toBeRebound;

        case 6:
          assert.strictEqual(rebindTo("asdf"), 42);
          _context.next = 9;
          return toBeRebound;

        case 9:
        case "end":
          return _context.stop();
      }
    }
  }, _marked[0], this);
}
```

and my (newly modified) code generates this code:
```
var toBeRebound = /*#__PURE__*/regeneratorRuntime.mark(function toBeRebound() {
  var originalValue;
  return regeneratorRuntime.wrap(function toBeRebound$(_context) {
    while (1) {
        case 0:
          originalValue = toBeRebound;
          _context.next = 3;
          return toBeRebound;

        case 3:
          assert.strictEqual(rebindTo(42), originalValue);
          _context.next = 6;
          return toBeRebound;

        case 6:
          assert.strictEqual(rebindTo("asdf"), 42);
          _context.next = 9;
          return toBeRebound;

        case 9:
        case "end":
          return _context.stop();
      }
    }
  }, toBeRebound, this);
});
function rebindTo(value) {
  var oldValue = toBeRebound;
  toBeRebound = value;
  return oldValue;
}
```
i think that `toBeRebound = value` call inside of `rebindTo` is changing `toBeRebound` in the main scope, but `function toBeRebound()` is still shadowing that value inside of the generator.

Andarist:
@modosc have you pushed out all changes? I wanted to look on your issue, but tests are passing for me on ur branch. Could u maybe specify how to reproduce ur situation?

packages/regenerator-transform/src/visit.js
Andarist: understanding all corner cases here is a little bit tricky and I have limited time now, but from what I understand about the issue:

- we cannot have the same identifier in the outer scope and in the inner scope because it would always act like the [safe example in the tests](https://github.com/facebook/regenerator/blob/9b6ebc0532ff373e06e3421b26800ac8979913d3/test/tests.es6.js#L1604-L1611)
- this leads me to the conclusion that the simplest answer for this is to just preserve transforming to the function declarations (like in the original transform) and not to function expression. What was the logic behind this change? 

It seems that the easiest would be to generate this:
```js
var toBeRebound$$ = /*#__PURE__*/regeneratorRuntime.mark(toBeRebound);
function toBeRebound() {
  var originalValue;
  return regeneratorRuntime.wrap(function toBeRebound$(_context) {
    while (1) {
        case 0:
          originalValue = toBeRebound;
          _context.next = 3;
          return toBeRebound;

        case 3:
          assert.strictEqual(rebindTo(42), originalValue);
          _context.next = 6;
          return toBeRebound;

        case 6:
          assert.strictEqual(rebindTo("asdf"), 42);
          _context.next = 9;
          return toBeRebound;

        case 9:
        case "end":
          return _context.stop();
      }
    }
  }, toBeRebound$$, this);
};
function rebindTo(value) {
  var oldValue = toBeRebound;
  toBeRebound = value;
  return oldValue;
}
```
This would:
- keep the `name` property untouched and make `Object.defineProperty` calls obsolete.
- keep proper identifiers in the proper scopes
- keep original reference (before rebinding) in scope so it can be accessed by the `regeneratorRuntime.wrap` call (which was previously accessed with `_marked[0]`)

modosc:
i opened a new pr (#309) based on the suggestion from @Andarist . the change is much simpler, tests pass, etc, so i'm going to close this one out.

PR #303, opened 2017-07-22
Setup a Lerna monorepo

### LINKED ISSUE #260: Use Lerna to manage packages/regenerator-* versions.
Babel uses [Lerna](https://lernajs.io/) to manage its own https://github.com/babel/babel/tree/master/packages directory, and I think it would be helpful for this repository, too. In particular, `lerna bootstrap` could probably replace [`link.sh`](https://github.com/facebook/regenerator/blob/master/link.sh) (recently introduced by 4af023c12a85073d8a3e6e315300d5f650745de7).

Andarist:
this fixes #260


.travis.yml
```diff
@@ -1,16 +1,15 @@
...
+  - npm run bootstrap
+
 language: node_js
 node_js:
   - "8"
   - "7"
   - "6"
   - "5"
   - "4"
-  - "0.12"
```
Andarist: those got removed as unsupported by lerna

packages/regenerator-preset/package.json
```diff
@@ -22,6 +22,6 @@
     "babel-plugin-transform-es2015-block-scoping": "^6.18.0",
     "babel-plugin-transform-es2015-classes": "^6.18.0",
     "babel-plugin-transform-es2015-for-of": "^6.18.0",
-    "regenerator-transform": "^0.9.3"
+    "regenerator-transform": "^0.9.12"
```
Andarist: dep updated to much `./packages/regenerator-transform`

package.json
```diff
@@ -25,21 +25,35 @@
...
   "scripts": {
-    "pretest": "./link.sh",
+    "bootstrap": "lerna bootstrap",
     "test": "node test/run.js"
   },
   "dependencies": {
     "babel-core": "^6.18.2",
+    "babel-plugin-syntax-async-functions": "^6.13.0",
+    "babel-plugin-syntax-async-generators": "^6.13.0",
+    "babel-plugin-transform-es2015-arrow-functions": "^6.8.0",
```
Andarist: all of those new deps here got hoisted manually by me from each package, it is supposed to make process of bootstrapping + hoisting faster (lerna)

lerna.json
```diff
@@ -0,0 +1,9 @@
+{
+  "lerna": "2.0.0",
+  "packages": [
+    "packages/*",
+    "."
```
Andarist: root directory could be removed and root's `package.json` marked as `"private": true` as this root "package" seems to be deprecated anyway

benjamn:
Awesome, glad to see this happen!

PR #305, opened 2017-07-22
Removed binding the invoke function to the possibly active domain

Andarist:
As discussed in - https://github.com/facebook/regenerator/pull/197#issuecomment-317135610


benjamn:
Sayonara `process.domain.bind`!

PR #306, opened 2017-07-23
fix build on node 8, update deps

modosc:
node 8 build fixes from #301 + update deps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ cd path/to/node_modules
 git clone git://github.com/facebook/regenerator.git
 cd regenerator
 npm install .
+npm run bootstrap
 npm test
 ```
 


benjamn:
Thanks for fixing this, @modosc! Does the merging of https://github.com/facebook/regenerator/pull/303 have any effect on the strategy of this PR? Hopefully this is the kind of thing that Lerna would help solve?

modosc:
no idea, I'll test it out locally. 

I found two separate issues - the first was the npm5 build failure, but the second only showed when i went back and forth between node <8 and node 8 on the same machine. The previous link.sh behavior left the local versions linked globally which required cleanup because of npm5's link behavior. Hopefully this pr isn't necessary anymore, I'll verify later today.

modosc:
ok - verified that `npm run bootstrap` fixes almost all of the issues i saw. still needed an `.npmrc` in the root directory or else a `package-lock.json` will be created on the initial `npm install`.

PR #307, opened 2017-07-23
Refactor/es modules

### LINKED ISSUE #302: [refactor] es modules, no global polution
**this would mean breaking change**

So basically what I would like to do (I can prepare PRs) is to refactor packages. Changes should be pretty straightforward and would include:
- including 2 distributions of `regenerator-runtime`, both commonjs and es
- stop auto pollution of global object - we can distribute also umd bundle if somebody needs it
- rewrite `regenerator-transform` to add an import for `regenerator-runtime` to the transformed module if needed - we can support old way (using global) with an option to the transform

Benefits:
- no more [weirdo code](https://github.com/facebook/regenerator/blob/83c5cec9a7945daa82e7431d0ab894d2058ef935/packages/regenerator-runtime/runtime-module.js) ;)
- tree-shakability of the `regenerator-runtime` - when one uses i.e. only normal generators then parts about async functions could be tree-shaken easily

Andarist:
This fixes #302

@benjamn 
So Im mostly done when it comes to refactoring `regenerator-runtime`.

The missing piece is to change `regenerator-transform` so it adds an import statement to the `regenerator-runtime` module. Im not strong with AST (although I know the basics) so an advice how this should be handled would be greatly appreciated! :)

Whats the best way to add a single import statement for a file with multiple generators?
How we differentiate what type of import (es/cjs) should be added?
Do we need to support a use case of including the regenerator itself in the outputted bundle? When?


benjamn:
Could we copy [the logic of `babel-plugin-transform-runtime`](https://github.com/babel/babel/blob/fe13ba8fc2ae0bb756cae6d980793ab0e4cf13f0/packages/babel-plugin-transform-runtime/src/index.js#L27)? Hopefully that should work even if `transform-runtime` is not otherwise being used.

Andarist:
@benjamn ive finished working on this, please take a look - aint sure if `regenerator-transform` adjustments were made in the best possible way, but its working 🎉 

Please review and let me know if there is anything I need to do to advance it further :)

Once this gets merged (and published) I'll try to update most notable projects depending on those packages.

Andarist:
@benjamn rebased against recent master

Andarist:
@benjamn any chances this could get reviewed? 😃

Unfortunately at the moment Im not sure why travis is failing, because locally everything is working fine. It would be great to debug it in ssh mode, but I dont have needed permissions to do it myself.

brettz9:
Could this be (rebased and) reviewed? With @babel/polyfill deprecated, there is no (direct) ESM replacement.

agilgur5:
@benjamn is there any chance you or someone from FB would be able to resurrect this? Don't mean to necro or force any work, but was hoping there might be a way to prioritize this especially as this is still under FB's org (and therefore theoretically supported by a very large company).
A lot of issues in the community end up pointing here to this PR that was never reviewed, so it doesn't seem as if there's much the community can do right now (other than fork) 😕 

In particular, as https://github.com/facebook/regenerator/issues/302#issuecomment-775211024 / https://github.com/babel/babel-polyfills/issues/59 point out, this PR would enable a "pure" polyfill. Currently all usage of `regenerator-runtime` causes pollution of global scope, which is not ideal for library authors.

PR #308, opened 2017-08-02
Configured prettier

Andarist:
Ive noticed in the other PR that there was some mismatch between PR code style and the desired style of the repo. This was also a slight problem for me when making changes as there were no linting rules around to help me and I had double check semicolon, quotes style etc - especially as I couldnt focus at one moment and was coming back to work several times when only I had some time.

This should help the situation - I've gone with the pure prettier as its super easy to add for the project and doesnt require too much configuration with its 'take it or leave it' approach.


package.json
```diff
@@ -23,9 +23,10 @@
     "url": "git://github.com/facebook/regenerator.git"
   },
   "main": "main.js",
-  "bin": "bin/regenerator",
+  "bin": "bin/regenerator.js",
```
Andarist: ive changed it so this file could also be 'caught' by the `lint-staged`'s `*.js` rule

PR #309, opened 2017-08-04
refactor mark declaration, add __PURE__ comments (#286)

modosc:
(see also #301 for old discussion)

with the current code, this:
```
export function* test() {}
export function* abc() {}
```
compiles to this:
```
var _marked = [test, abc].map(regeneratorRuntime.mark);

export function test() {
  return regeneratorRuntime.wrap(function test$(_context) {
    while (1) {
        case 0:
        case "end":
          return _context.stop();
      }
    }
  }, _marked[0], this);
}

export function abc() {
  return regeneratorRuntime.wrap(function abc$(_context2) {
    while (1) {
      switch (_context2.prev = _context2.next) {
        case 0:
        case "end":
          return _context2.stop();
      }
    }
  }, _marked[1], this);
}
```
because of the `[test, abc].map` call `uglifyjs` won't strip out `test` or `abc` if they're never actually imported (`webpack` does correctly identify them as non-imported though). 

this pr changes two things:
1. the `map` call is unrolled and each call to `regeneratorRuntime.mark` has the magic `#__PURE__` comment added to it so that uglify knows it can drop the code:
```
var _marked = /*#__PURE__*/regeneratorRuntime.mark(test),
    _marked2 = /*#__PURE__*/regeneratorRuntime.mark(abc);
export function test() {
  return regeneratorRuntime.wrap(function test$(_context) {
    while (1) {
        case 0:
        case "end":
          return _context.stop();
      }
    }
  }, _marked, this);
}
export function abc() {
  return regeneratorRuntime.wrap(function abc$(_context2) {
    while (1) {
      switch (_context2.prev = _context2.next) {
        case 0:
        case "end":
          return _context2.stop();
      }
    }
  }, _marked2, this);
}
```
2. the magic comment is also added to function expressions, so this:
```
var foo = function* (){}
```
is compiled to this:
```
var foo = /*#__PURE__*/regeneratorRuntime.mark(function _callee() {
    while (1) {
        case 0:
        case "end":
          return _context.stop();
      }
    }
  }, _callee, this);
});
```

i added tests to verify the new variable declaration scheme. i haven't been able to get `recast` to return comments to me in a test though - i've verified that this works via `bin/regenerator` (to verify the comments are there) and by using it in my project with `uglifyjs` and verifying that unused generators are being stripped out.

### DIFF
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -153,6 +153,7 @@ exports.visitor = {
 
       if (wasGeneratorFunction && t.isExpression(node)) {
         util.replaceWithOrRemove(path, t.callExpression(util.runtimeProperty("mark"), [node]))
+        path.addComment("leading", "#__PURE__");
       }
 
       // Generators are processed in 'exit' handlers so that regenerator only has to run on
@@ -188,18 +189,34 @@ function getOuterFnExpr(funPath) {
     }
 
     let markDecl = getRuntimeMarkDecl(pp);
-    let markedArray = markDecl.declarations[0].id;
-    let funDeclIdArray = markDecl.declarations[0].init.callee.object;
-    t.assertArrayExpression(funDeclIdArray);
+    let declarations = markDecl.declarations;
+    let index = declarations.length;
 
-    let index = funDeclIdArray.elements.length;
-    funDeclIdArray.elements.push(node.id);
+    // this can change from pass to pass if something else is altering our body
+    // so we have to look it up every time :/
+    let declarationsIndex = pp.node.body.findIndex((p) => {
+      return t.isVariableDeclaration(p) && p.declarations === declarations;
+    });
+    assert(!isNaN(declarationsIndex));
 
-    return t.memberExpression(
-      markedArray,
-      t.numericLiteral(index),
-      true
-    );
+    // get a new unique id for our marked variable
+    let markedId = pp.scope.generateUidIdentifier("marked");
+
+    // now push our new declaration into our list
+    declarations.push(t.variableDeclarator(
+      markedId,
+      t.callExpression(util.runtimeProperty("mark"), [node.id])
+    ));
+
+    // assemble the path to our mark declarations
+    let bodyPathName = 'body.' + String(declarationsIndex) + '.declarations.' + String(index) + '.init';
+    let bodyPath = pp.get(bodyPathName);
+
+    // mark our call to util.runtimeProperty("mark") as pure to enable tree-shaking
+    bodyPath.addComment("leading", "#__PURE__");
+
+    // and return our new id
+    return markedId;
   }
 
   return node.id;
@@ -214,20 +231,7 @@ function getRuntimeMarkDecl(blockPath) {
     return info.decl;
   }
 
-  info.decl = t.variableDeclaration("var", [
-    t.variableDeclarator(
-      blockPath.scope.generateUidIdentifier("marked"),
-      t.callExpression(
-        t.memberExpression(
-          t.arrayExpression([]),
-          t.identifier("map"),
-          false
-        ),
-        [util.runtimeProperty("mark")]
-      )
-    )
-  ]);
-
+  info.decl = t.variableDeclaration("var", [])
   blockPath.unshiftContainer("body", info.decl);
 
   return info.decl;
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -8,14 +8,14 @@
  * the same directory.
  */
 
+"use strict";
+
 import assert from "assert";
 import * as t from "babel-types";
 import { hoist } from "./hoist";
 import { Emitter } from "./emit";
 import replaceShorthandObjectMethod from "./replaceShorthandObjectMethod";
 import * as util from "./util";
-
-let getMarkInfo = require("private").makeAccessor();
 
 exports.name = "regenerator-transform";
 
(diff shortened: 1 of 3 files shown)


packages/regenerator-transform/src/visit.js
```diff
@@ -214,20 +231,7 @@ function getRuntimeMarkDecl(blockPath) {
...
-          t.arrayExpression([]),
-          t.identifier("map"),
-          false
-        ),
-        [util.runtimeProperty("mark")]
-      )
-    )
-  ]);
-
+  info.decl = t.variableDeclaration("var", [])
```
Andarist: this looks odd, should it stay here in the code?

test/tests.transform.js
```diff
@@ -47,3 +47,57 @@ describe("_blockHoist nodes", function() {
...
+    const code = recast.print(transform(ast)).code;
+    assert.notStrictEqual(
+      code.indexOf(
+        'var _marked = regeneratorRuntime.mark(foo), _marked2 = regeneratorRuntime.mark(bar);'
+      ), -1
+    );
+  });
+});
+
+// TODO - how do i get recast to return comments?
```
Andarist: i aint sure how what u are trying to achieve here exactly, ive managed to access comments like this:
```
recast.parse('var _marked = /*#__PURE__*/regeneratorRuntime.mark(test)').program.body[0].declarations[0].init.comments
```
maybe that helps you somehow

test/tests.transform.js
```diff
@@ -47,3 +47,57 @@ describe("_blockHoist nodes", function() {
...
+    );
+  });
+  xit("should work with a function declaration", function() {
+    var ast = recast.parse('function* foo(){};', {
+      parser: require("babylon")
+    });
+    const code = recast.print(transform(ast)).code;
+    assert.notStrictEqual(
+      code.indexOf(
+        'var a = /*#__PURE__*/Object.defineProperty( /*#__PURE__*/regeneratorRuntime.mark'
```
Andarist: why the `Object.defineProperty` call here? didnt we ditch it?

packages/regenerator-transform/src/visit.js
```diff
@@ -188,18 +189,34 @@ function getOuterFnExpr(funPath) {
...
-      markedArray,
-      t.numericLiteral(index),
-      true
-    );
+    let declarations = markDecl.declarations;
+    let index = declarations.length;
+
+    // this can change from pass to pass if something else is altering our body
+    // so we have to look it up every time :/
+    let declarationsIndex = pp.node.body.findIndex((p) => {
```
Andarist: maybe we could gather those declarations in context instead of inserting them right away and insert them all at once in program's exit visitor? 

aint sure if thats better, just an idea

test/tests.transform.js
modosc: sorry, copy pasta. will fix.

packages/regenerator-transform/src/visit.js
modosc: it's basically the same functionality as before which was to memoize the place the declarations were to go - i guess since it's so much smaller now it could go in the caller insetad?

packages/regenerator-transform/src/visit.js
modosc: happy to look into this, it's probably best to not have to traverse the tree downward every time. i'm new to babel/ast stuff so it may take a bit to sort out.

test/tests.transform.js
modosc: just updated all the tests, should be clearer now.

packages/regenerator-transform/src/visit.js
Andarist: im not a babel/ast expert either :P so i dont know what best practices are

just had such an idea how it should be possible to address the issue from the comment

test/tests.transform.js
```diff
@@ -47,3 +47,102 @@ describe("_blockHoist nodes", function() {
     assert.deepEqual(names, ["hoistMe", "doNotHoistMe", "oyez"]);
   });
 });
+
+context("functions", function() {
```
Andarist: are those good tests? I mean - they are highly coupled to the internal implementation, shouldnt we just test if the outcome is treeshakeable?

modosc: i'll add an uglify test.

modosc: done.

Andarist: 👏 great work!

modosc: hah, thanks. is there anything else i can do to help expedite this merge? it makes a huge difference for me (and probably anyone else using `redux-sagas`).

benjamn: I'll try to look at this today! Thanks so much to both of you for finding a simpler solution and perfecting it.

Andarist: @modosc as maintainer of the redux-saga im a little bit surprised that this make u a huge difference. I thought this is more 'nice to have' and 'always better have this in place' rather than a game changer. Do u have any numbers for ur project? What will be tree shaken after this change?

I see when this might make a change when u have some shared saga utils module used between projects, but i think most dont

modosc: @Andarist lol, didn't realize that. 

in our case we're generating api sagas automatically from swagger docs and grouping them by model. we've got at least one saga for every single version / endpoint / method so for some models we have >30 sagas in a single file. most of them aren't used but due to this issue using a single saga from a file pulls in all the others.

in retrospect i probably overstated the value of this to the average user - in normal circumstances i can't imagine having that many sagas sitting around that aren't actually used (although i can imagine situations with code-spitting where you end up bundling an entire module because you want access to a single saga inside of it).

Andarist: > in our case we're generating api sagas automatically from swagger docs and grouping them by model.

metaprogramming level hard :D would love to see this in action

> although i can imagine situations with code-spitting where you end up bundling an entire module because you want access to a single saga inside of it

yeah, thats true, however i see the issue only for 'util' saga files, others are rather grouped by some kind of functionality and they rather work only together

packages/regenerator-transform/src/visit.js
benjamn: I pushed another commit that contains some improvements for this logic.

benjamn:
Thanks @modosc and @Andarist for iterating on this and making it so much simpler!

Now to publish these packages with Lerna for the first time…

benjamn:
You can see these changes live in the sandbox now! http://facebook.github.io/regenerator/

Andarist:
wohoooooo! 🎉

PR #310, opened 2017-08-15
Breaking: remove transform-runtime

hzoo:
ideally would update the packages to v7.0 alpha too but can do that later I guess?

and do we need private if we have `Symbol`?


hzoo:
Are we ok with doing another bump to drop node 0.10/0.12 by doing this?

PR #311, opened 2017-08-16
Update visit.js

hzoo:
remove the name, causing an issue with https://github.com/babel/babel/issues/6126 in 6.x for some version (it might be the babel-core version though)

### DIFF
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -17,8 +17,6 @@ import { Emitter } from "./emit";
 import replaceShorthandObjectMethod from "./replaceShorthandObjectMethod";
 import * as util from "./util";
 
-exports.name = "regenerator-transform";
-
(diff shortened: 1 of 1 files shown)


hzoo:
Ok nvm, we don't have to do this.

loganfsmyth:
Sorry I missed that this didn't get merged, we definitely need to undo this. It's a semver-breaking change that make this plugin fail on every version of `babel-core` except apparently the most recent version.

loganfsmyth:
Hmm actually I guess since you bumped to 0.10 it's not semver-breaking for regnenerator, my mistake. It does mean we can't use 0.10 on 6.x though.

benjamn:
Can we detect the Babel version and add the `name` conditionally somehow?

loganfsmyth:
Yup. Let's close this PR and I'll open a new one.

PR #312, opened 2017-08-17
Only export plugin name for Babel>=7.x

loganfsmyth:
Alternate approach for https://github.com/facebook/regenerator/pull/311 that will avoid breaking changes for Babel 6.x

### DIFF
--- a/packages/regenerator-transform/src/index.js
+++ b/packages/regenerator-transform/src/index.js
@@ -8,6 +8,21 @@
  * the same directory.
  */
 
-export default function () {
-  return require("./visit");
+export default function (context) {
+  const plugin = {
+    visitor: require("./visit").visitor,
+  };
+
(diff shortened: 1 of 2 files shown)


packages/regenerator-transform/src/index.js
```diff
@@ -8,6 +8,21 @@
...
+  };
+
+  // Some presets manually call child presets, but fail to pass along the
+  // context object. Out of an abundance of caution, we verify that it
+  // exists first to avoid causing unnecessary breaking changes.
+  const version = context && context.version;
+
+  // The "name" property is not allowed in older versions of Babel (6.x)
+  // and will cause the plugin validator to throw an exception.
+  if (version && parseInt(version) >= 7) {
```
loganfsmyth: No radix, groooss

PR #313, opened 2017-09-07
Don't error if Gp[toStringTagSymbol] is read-only

JakeChampion:
I'm not so sure that this will be merged. I've noticed that if using https://polyfill.io with this project that this line throws an error because the IteratorPrototype[Symbol.toString] is read-only


JakeChampion:
Signed!

zloirock:
So maybe it should be fixed in `polyfill.io`?

JakeChampion:
@zloirock Good point, will push a fix there 👍

PR #319, opened 2017-10-05
use babel-types v7

hzoo:
Fixes https://github.com/babel/babel/issues/6262

### DIFF
--- a/packages/regenerator-transform/package.json
+++ b/packages/regenerator-transform/package.json
@@ -29,7 +29,7 @@
     ]
   },
   "dependencies": {
-    "babel-types": "7.0.0-beta.2",
+    "babel-types": "7.0.0-beta.3",
     "private": "^0.1.6"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


hzoo:
cc @benjamn for https://github.com/babel/babel/issues/6262

damassi:
Any update on merging this?

hzoo:
This needs an update to peerDep on @babel/core and then just use that version instead of babel-types under a beta version.

@benjamn not sure what to do here because this also is causing bugs on our side because of the different versions of babel-types people are getting (also a bigger bundle in babel-standalone).

packages/regenerator-transform/package.json
```diff
@@ -29,7 +29,7 @@
     ]
   },
   "dependencies": {
-    "babel-types": "^6.19.0",
+    "babel-types": "7.0.0-beta.2",
```
benjamn: This should be `-beta.3` now, right?

damassi:
Thanks @benjamn 👍

hzoo:
Sorry this should of been `@babel/types` `v7.0.0-beta.32` now 😛

PR #329, opened 2017-11-25
Use plugin-provided context.types instead of importing babel-types.

### LINKED ISSUE #328: Remove `babel-types` dependency and use the one passed to the plugin instead
So we don't have to keep updating `babel-types` and having incompatibilities with v7.

We'd need to remove the `import * as t from 'babel-types'` in each file in https://github.com/facebook/regenerator/tree/master/packages/regenerator-transform/src

https://github.com/facebook/regenerator/blob/e0d0575d764503cf2817da2c553fd3065c42848f/packages/regenerator-transform/src/index.js#L11-L13

Here we have to pass `context.types` as `t` to all files somehow (babel-types is passed so each babel plugin already)? Not sure the best way to do this other than wrapping files in a function with the passed `t`

benjamn:
Fixes #328.

cc @hzoo @loganfsmyth 

Instead of passing the `context.types` object throughout `regenerator-transform` (really invasive), or wrapping every module with some sort of constructor function (which would require abandoning both `require` and `import`), I decided to use stack-based local state in the `util.js` module, so that `currentTypes` can be retrieved from anywhere, as long as there's a function wrapped with `util.wrapWithTypes` currently on the stack. Since `regenerator-transform` is entirely synchronous, this turns out to be an easy property to enforce, with only one call to `util.wrapWithTypes`.

### DIFF
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -11,15 +11,14 @@
 "use strict";
 
 import assert from "assert";
-import * as t from "babel-types";
 import { hoist } from "./hoist";
 import { Emitter } from "./emit";
 import replaceShorthandObjectMethod from "./replaceShorthandObjectMethod";
 import * as util from "./util";
 
-exports.visitor = {
+exports.getVisitor = ({ types: t }) => ({
   Function: {
-    exit: function(path, state) {
+    exit: util.wrapWithTypes(t, function(path, state) {
       let node = path.node;
 
       if (node.generator) {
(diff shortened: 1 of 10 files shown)


packages/regenerator-transform/src/visit.js
```diff
@@ -11,15 +11,14 @@
...
 import { hoist } from "./hoist";
 import { Emitter } from "./emit";
 import replaceShorthandObjectMethod from "./replaceShorthandObjectMethod";
 import * as util from "./util";
 
-exports.visitor = {
+exports.getVisitor = ({ types: t }) => ({
   Function: {
-    exit: function(path, state) {
+    exit: util.wrapWithTypes(t, function(path, state) {
```
benjamn: Everything `regenerator-transform` does (or, more specifically, all code that calls `util.getTypes()`) happens within this `exit` method, so this is the only function that we need to wrap.

loganfsmyth:
Seems like a solid solution to me, good call.

Eventually, once enough of the ecosystem is transitioned to Babel 7.x that plugins are willing to add explicit `peerDependencies` on `@babel/core@7`, at least it'll be easy to change to 

```
import { types as t } from "@babel/core";
```
but I do think it's too early for that since it'd block the ability to support both Babel versions in parallel.

hzoo:
Can submit a PR to Babel after this

PR #330, opened 2017-11-26
Fix hoisting in combination with arrow functions

TimWolla:
see babel/babel#6908


benjamn:
Good catch @TimWolla!

PR #335, opened 2017-12-30
Fix self rebind in nested function

### LINKED ISSUE #225: Nested functions can't be properly rebound
The following code does not transform as expected:

``` js
function parent1() {
  function toBeRebound() {
    toBeRebound = 42;
  }
  toBeRebound();
  console.log(toBeRebound);
}

parent1(); // 42

function *parent2() {
  function toBeRebound() {
    toBeRebound = 42;
  }
  toBeRebound();
  console.log(toBeRebound);
}

parent2().next(); // function toBeRebound() { ... }
```

Functions that are nested in generators are transformed into `originalName = function originalName() { ... }` function expressions. In this case, function name can't be rebound from the inside of the function as such identifiers are bound to function expression and not to variable.

bsuh:
Fixes #225

### DIFF
--- a/packages/regenerator-transform/src/hoist.js
+++ b/packages/regenerator-transform/src/hoist.js
@@ -87,7 +87,7 @@ exports.hoist = function(funPath) {
           "=",
           node.id,
           t.functionExpression(
-            null,
+            path.scope.generateUidIdentifierBasedOnNode(node),
             node.params,
             node.body,
             node.generator,
--- a/packages/regenerator-transform/src/hoist.js
+++ b/packages/regenerator-transform/src/hoist.js
@@ -87,7 +87,7 @@ exports.hoist = function(funPath) {
           "=",
           node.id,
           t.functionExpression(
-            node.id,
+            null,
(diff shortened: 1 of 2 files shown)


packages/regenerator-transform/src/hoist.js
```diff
@@ -87,7 +87,7 @@ exports.hoist = function(funPath) {
           "=",
           node.id,
           t.functionExpression(
-            node.id,
+            null,
```
nicolo-ribaudo: Can you use `path.scope.generateUidIdentifierBasedOnNode(node)`? By doing this, we give a name to the function (so it helps during debugging) and we avoid shadowing the original variable.

benjamn:
Thanks for this fix @bsuh and thanks for the review @nicolo-ribaudo!

benjamn:
Published to npm as `regenerator-transform@0.12.3`: https://github.com/facebook/regenerator/commit/58c0eca04145da79c263194a605ca5a6294a72b4

PR #339, opened 2018-01-19
add outerFn attribute to generator object

greenplumm:


greenplumm:
Please add the outerFn attribute of the generator object so that you will know the name of the function as soon as the generator function executes

PR #344, opened 2018-05-22
Preserve the location of the replaced 'yield' nodes.

### LINKED ISSUE #342: AST location information lost when emitting yield replacement (and others)
See in this sample from http://sokra.github.io/source-map-visualization for a sample async function passed through Babel.
screen shot 2018-02-16 at 10 46 43 am

Note how the `return` before `doAsync` is blue instead of white, and the right side shows no generated mapping for the `await`.

This specific example is for https://github.com/facebook/regenerator/blob/58c0eca04145da79c263194a605ca5a6294a72b4/packages/regenerator-transform/src/emit.js#L1189

Where essentially ideally it would do
```
let ret = t.returnStatement(arg || null);
ret.loc = expr.loc;
self.emit(ret);
```
so the location information from the yield expression is carried over onto the `return` statement.

Depending how ambitious we want to be, this probably also comes up in a few other places where regenerator creates new AST nodes, like all of these cases: https://github.com/facebook/regenerator/blob/58c0eca04145da79c263194a605ca5a6294a72b4/packages/regenerator-transform/src/emit.js#L1162
the replacement UpdateExpression will not have the location information that the original node did, which may not be ideal.

Think I've got someone who wants to look into making a PR for this. Handling the `return` at least would be ideal, but it might be good to account for the other cases too.

loganfsmyth:
Fixes #342 
Refs https://github.com/devtools-html/debugger.html/issues/5561, https://github.com/devtools-html/debugger.html/issues/6400

It doesn't look like there are any tests that validate the source mappings, but if you feel strongly about it I'd be happy to explore adding some.

Like I mentioned in #342, stuff like https://github.com/facebook/regenerator/blob/58c0eca04145da79c263194a605ca5a6294a72b4/packages/regenerator-transform/src/emit.js#L1162 also loses the location because it replaces the nodes with fresh ones that don't have a location, but I'm leaving that for now since it's a more aggressive change.

### DIFF
--- a/packages/regenerator-transform/src/emit.js
+++ b/packages/regenerator-transform/src/emit.js
@@ -1172,21 +1172,29 @@ Ep.explodeExpression = function(path, ignoreResult) {
     if (arg && expr.delegate) {
       let result = self.makeTempVar();
 
-      self.emit(t.returnStatement(t.callExpression(
-        self.contextProperty("delegateYield"), [
+      let ret = t.returnStatement(t.callExpression(
+        self.contextProperty("delegateYield"),
+        [
           arg,
           t.stringLiteral(result.property.name),
           after
         ]
-      )));
+      ));
+      ret.loc = expr.loc;
 
+      self.emit(ret);
       self.mark(after);
 
       return result;
     }
 
     self.emitAssign(self.contextProperty("next"), after);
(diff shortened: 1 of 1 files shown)


loganfsmyth:
Thanks!

PR #346, opened 2018-06-03
Fix unsafe-eval CSP violation - closes #336

### LINKED ISSUE #336: `Function("...")` causes unsafe-eval CSP violation in Chrome extension
I'm trying to use [react-form](https://github.com/react-tools/react-form) in my Chrome extension options page. It has `regenerator-runtime` as the transitive dependency that throws the error, so I figured I'll create an issue here.

Thanks for your time! I appreciate your support of the JS community.

## Expected Behavior

`regenerator-runtime` works silently like a good boy.

## Actual Behavior

```js
require("regenerator-runtime") // Uncaught EvalError
```

Offending line: `node_modules/regenerator-runtime/runtime-module.js:10`

```js
var g = (function() { return this })() || Function("return this")();
```

Version: `regenerator-runtime@0.11.1`

Full error:

> Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' blob: filesystem: chrome-extension-resource:".

## Environment

Chrome extension Options page, Chrome 63.0.3239.132 (64-bit), Windows 10

Benjamin-Dobell:
Refer to #336 for details.

### DIFF
--- a/packages/regenerator-runtime/runtime-module.js
+++ b/packages/regenerator-runtime/runtime-module.js
@@ -7,7 +7,9 @@
 
 // This method of obtaining a reference to the global object needs to be
 // kept identical to the way it is obtained in runtime.js
-var g = (function() { return this })() || Function("return this")();
+var g = (function() {
+  return this ? this : (typeof self !== 'undefined' ? self : undefined)
+})() || Function("return this")();
 
 // Use `getOwnPropertyNames` because not all browsers support calling
 // `hasOwnProperty` on the global `self` object in a worker. See #183.
--- a/packages/regenerator-runtime/runtime-module.js
+++ b/packages/regenerator-runtime/runtime-module.js
@@ -8,7 +8,7 @@
 // This method of obtaining a reference to the global object needs to be
 // kept identical to the way it is obtained in runtime.js
 var g = (function() {
-  return this ? this : (typeof self !== 'undefined' ? self : undefined)
+  return this || (typeof self === "object" && self);
 })() || Function("return this")();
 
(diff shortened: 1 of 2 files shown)


Benjamin-Dobell:
I think there's something slightly wrong with the Travis CI setup for this project. I just rebased on master in hope of getting the all clear, but all NPM versions failed in the same fashion, running `bootstrap`.

I can recreate, and remedy this, this locally from a fresh clone. Specifically, the issue seems to be that we need to install the `node_modules` of the `packages/`. The following gist demonstrates the problem and solution:

https://gist.githubusercontent.com/Benjamin-Dobell/cf86ce36fe8bf6ddfa06ada47725e44b/raw/6841c94b9758ed973ccced72102a9835cfc764ca/gistfile1.txt

Nonetheless, I'd appreciate it if this could PR could be reviewed & merged 😄

torarvid:
At my office, we have been waiting for this PR to be merged for over a month. I’m thinking we have to fork the project soon to have this CSP issue fixed, which seems like a shame..

JakeChampion:
I think we should use the same approach as core-js for this -- https://github.com/zloirock/core-js/blob/580eef389f9be01a00ce32f550fea6cc6296df28/packages/core-js/internals/global.js#L1-L5

benjamn:
These changes have been published as `regenerator-runtime@0.12.1`, though I would also request your 👀 on #353, which attempts to remove the need for hacks to discover the global object.

PR #347, opened 2018-06-18
Avoid reuse of AST nodes by making defensive clones

Andarist:
**Motivation**
Duplicated AST nodes were causing some obscure, hard to find bugs and they got forbidden in [babel's output](https://github.com/babel/babel/pull/7149).

At the moment this is unfortunately only [checked in core tests](https://github.com/babel/babel/blob/d383659ca6adec54b6054f77cdaa16da88e8a171/packages/babel-helper-transform-fixture-test-runner/src/index.js#L128). Because regenerator is used by babel as core transform and because it produces numerous duplicated nodes (only 1 "kind" of them intentionally) some nasty lines got introduced to babel's check functions to allow regenerator-created nodes.

I'd like to get rid of those nasty lines with this PR 😉

### DIFF
--- a/packages/regenerator-transform/src/emit.js
+++ b/packages/regenerator-transform/src/emit.js
@@ -34,9 +34,11 @@ function Emitter(contextId) {
   // that have been marked as branch/jump targets.
   this.marked = [true];
 
+  this.insertedLocs = new Set();
+
   // The last location will be marked when this.getDispatchLoop is
   // called.
-  this.finalLoc = loc();
+  this.finalLoc = this.loc();
 
   // A list of all leap.TryEntry statements emitted.
   this.tryEntries = [];
@@ -55,8 +57,18 @@ exports.Emitter = Emitter;
 // the amazingly convenient benefit of allowing the exact value of the
 // location to be determined at any time, even after generating code that
 // refers to the location.
-function loc() {
-  return util.getTypes().numericLiteral(-1);
+Ep.loc = function() {
+  const l = util.getTypes().numericLiteral(-1)
+  this.insertedLocs.add(l);
+  return l;
+}
+
+Ep.getInsertedLocs = function() {
+  return this.insertedLocs;
+}
+
+Ep.getContextId = function() {
+  return util.getTypes().clone(this.contextId);
 }
 
 // Sets the exact value of the given location to the offset of the next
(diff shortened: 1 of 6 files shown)


packages/regenerator-transform/src/emit.js
```diff
@@ -34,9 +34,11 @@ function Emitter(contextId) {
   // that have been marked as branch/jump targets.
   this.marked = [true];
 
+  this.insertedLocs = new Set();
```
Andarist: not sure what version of node do u want to support here, should I just leave it or import a ponyfill?

packages/regenerator-transform/src/visit.js
```diff
@@ -150,6 +150,18 @@ exports.getVisitor = ({ types: t }) => ({
         path.addComment("leading", "#__PURE__");
       }
 
+      const insertedLocs = emitter.getInsertedLocs();
+
+      path.traverse({
```
Andarist: this is a hack-ish solution for "intentional" duplicates, cant rly clone them when inserting to AST because they need to be the same object for [`mark`](https://github.com/facebook/regenerator/blob/92f5f8534d5005cf56d894a782e12b842628b5ed/packages/regenerator-transform/src/emit.js#L68) to take effect on all inserted instances

This whole thing with updating those nodes after they get inserted simplifies the transform. Rewriting it to avoid duplicates has too big cost and would be really hard.

I've considered tracking all insertions but it would complicate things quite a bit because there is no single place for AST insertions (kinda general rule for any babel transform 😉 )

So I've decided to just "fix" the transformed subtree after whole transform takes place.

benjamn:
Your choice of wording is interesting: the problem with these nodes is that they're *not* duplicated, but rather reused in multiple places in the AST. I'm not opposed to your solution, but calling `t.cloneDeep` is introducing new duplication where previously there was none. 🤔

benjamn:
> So I've decided to just "fix" the transformed subtree after whole transform takes place.

I support this 100%.

Andarist:
Your wording is indeed more accurate 😉 

And you are ofc right that probably from time to time those clones are defensive, although I've inserted them only at places that were causing nodes reuse.

> I'm not opposed to your solution, but calling t.cloneDeep is introducing new duplication where previously there was none.

This might be somewhat less memory efficient but ultimately aligns the codebase to what babel's monorepo is doing at the moment.

Im not aware of any issue fixed by those changes here, but we can imagine other transforms relaying on nodes being unique and i.e. doing some caching or storing some metadata - those cases were breaking from time to time for babel users and also for babel's codebase because of nodes reuse.

benjamn:
@Andarist Do you think this should be a minor semantic version bump, or just a patch?

packages/regenerator-transform/src/emit.js
benjamn: I'm comfortable with only supporting Node 4+.

PR #352, opened 2018-08-02
Replace unreliable Lerna bootstrapping with a simple test/run.sh script.

benjamn:
With any luck, simply updating to the latest (release candidate) version of Lerna will fix the problems reported by @Benjamin-Dobell [here](https://github.com/facebook/regenerator/pull/346#issuecomment-406909615). Properly linking local packages is definitely Lerna's job, and there's nothing particularly unusual about the organization of this repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 before_install:
-  - npm install lerna@2.11.0
-
-install:
-  - npm run bootstrap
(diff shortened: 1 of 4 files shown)


benjamn:
Well, that didn't work. If the latest Lerna can't get this right, I think it's time to write a dumb script that does what we need.

PR #353, opened 2018-08-03
Convert runtime.js to a normal CommonJS module.

benjamn:
If this works, then it removes the need for hacks to discover the global object, which continues to be problematic (e.g. #336, #346).

Importantly, the `regenerator-runtime/runtime.js` module can still be evaluated as a non-CJS script, and it will not pollute the global scope except for defining `regeneratorRuntime`.


benjamn:
@Benjamin-Dobell Thoughts on this approach?

MattiasBuelens:
Any chance you could push out a patch release for this?

wmertens:
So - I think this caused my build to fail. Is it Babel's responsibility to make this global?

MattiasBuelens:
Thanks for the release! 🙂

Since this is a minor bump on a 0.x release, packages such as `@babel/polyfill` and `@babel/runtime` will need to update their dependency versions. Are you going to make a PR for updating Babel, or is it okay if I submit one?

MattiasBuelens:
I've opened babel/babel#9442 to update `@babel/polyfill`, `@babel/runtime` and `@babel/runtime-corejs2` to the latest version of the runtime.

PR #356, opened 2018-09-21
 Add "importRuntime" option to "transform-runtime"

nicolo-ribaudo:
> When "importRuntime" is true, regenerator doesn't assume that
> `regeneratorRuntime` is available globally, but it injects an `import`
> statement (or `require`, depending on the source type)

Example output:
```js
var _regeneratorRuntime = require("regenerator-runtime");

var _marked =
/*#__PURE__*/
_regeneratorRuntime.mark(f);

function f() {
  return _regeneratorRuntime.wrap(function f$(_context) {
      case 0:
      case "end":
       return _context.stop();
    }
  }, _marked, this);
}
```

I had to update the Babel deps because I hit https://github.com/babel/babel/issues/8508 and https://github.com/babel/babel/issues/8722.


package.json
```diff
@@ -28,21 +28,21 @@
...
+    "@babel/types": "^7.0.0",
     "commoner": "^0.10.8",
     "private": "^0.1.6",
     "recast": "^0.15.3",
-    "regenerator-preset": "^0.12.0",
-    "regenerator-runtime": "^0.12.1",
-    "regenerator-transform": "^0.13.3",
+    "regenerator-preset": "file:packages/regenerator-preset",
+    "regenerator-runtime": "file:packages/regenerator-runtime",
+    "regenerator-transform": "file:packages/regenerator-transform",
```
nicolo-ribaudo: This changes were made by lerna I think.

packages/regenerator-transform/src/util.js
```diff
@@ -23,10 +25,30 @@ export function getTypes() {
   return currentTypes;
 }
 
-export function runtimeProperty(name) {
+const runtimeNames = new WeakMap();
+
+export function runtimeProperty(name, scope, opts) {
```
benjamn: Now that `runtimeProperty` takes options and does more than just construct a member expression, I think we should move it out of `util.js` and make it a method of the `Emitter` class. That way we won't need to pass the options explicitly, since they're stored as a property of the `Emitter` object now. From what I can tell, everywhere `runtimeProperty` was previously called, we have access to an `Emitter` instance, so this change should be pretty straightforward. What do you think @nicolo-ribaudo?

benjamn (changes requested):
This seems like the right direction, though I have one substantive suggestion.

package.json
benjamn: Should we revert the lerna changes?

benjamn (approved):

loganfsmyth:
To throw it out there, I don't have a problem with landing this if you want, but I don't think it's something we'd recommend people use on the Babel side since `plugin-transform-runtime` also handles injecting a reference to the runtime and is already very recommended in Babel 7.

nicolo-ribaudo:
Closing since this is now handled by `@babel/plugin-transform-runtime`

PR #362, opened 2018-11-13
Upgrade browserify to get past CVE

zpao:
https://nvd.nist.gov/vuln/detail/CVE-2018-16472 claims
cached-path-relative is potentially vulnerable to an exploit. As
a transitive dependency, the risks are virtually non-existant, but might
as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -310,9 +310,9 @@
       }
     },
     "JSONStream": {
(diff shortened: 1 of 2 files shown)


benjamn:
Thanks @zpao! Looks like tests are failing on `master` right now, so I'll fix that and then merge this.

PR #364, opened 2019-01-25
Replace "require" with "import" in regenerator-transform source

salamagd:
I think @cito125 (`mad_babeler` on BabelJS Slack) has already been in contact with you about some issues we're having using `regenerator-transform` together with Babel 7, React Native and `proxyquire`. Turns out that replacing the `require` statements in the source with ES module `import` statements solves our issue, as we no longer end up importing the `emit.js` source twice (which causes issues with the type assertion on line 131 of `leap.js`.

Hopefully this is a pretty non-controversial change to update all the `require` statements in the package with `import`s instead!

### DIFF
--- a/packages/regenerator-transform/src/index.js
+++ b/packages/regenerator-transform/src/index.js
@@ -5,9 +5,11 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import { getVisitor } from "./visit";
+
 export default function (context) {
   const plugin = {
-    visitor: require("./visit").getVisitor(context),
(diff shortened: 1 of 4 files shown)


cito125:
Yup tis I the `mad_babeler`! There seems to be some conflict with [proxyquire](https://github.com/thlorenz/proxyquire) used by us and [pirates](https://github.com/ariporad/pirates) as used by both `babel-preset-react-native` and `metro-react-native-babel-preset` in how and/or when `require` is transpiled and/or mocked.

PR #365, opened 2019-02-12
Don't capture "this" when not needed

nicolo-ribaudo:
Fixes babel/babel#4056

As explained in that issue, regenerator prevents engines from collectiong `this` even when it is not used, causing a memory leak.

Input:
```js
function f() {
  return async _ => {};
}

function g() {
  return async _ => this;
}
```

<details>
<summary>Output before this PR:</summary>

```js
function f() {
  return _ => {
    return regeneratorRuntime.async(function _callee$(_context) {
        case 0:
        case "end":
          return _context.stop();
      }
    }, null, this);
  };
}

function g() {
  return _ => {
    return regeneratorRuntime.async(function _callee2$(_context2) {
      while (1) switch (_context2.prev = _context2.next) {
        case 0:
          return _context2.abrupt("return", this);

        case 1:
        case "end":
          return _context2.stop();
      }
    }, null, this);
  };
}
```
</details>

<details>
<summary>Output after this PR:</summary>

```js
function f() {
  return _ => {
    return regeneratorRuntime.async(function _callee$(_context) {
        case 0:
        case "end":
          return _context.stop();
      }
    }); // <-- difference here!
  };
}

function g() {
  return _ => {
    return regeneratorRuntime.async(function _callee2$(_context2) {
      while (1) switch (_context2.prev = _context2.next) {
        case 0:
          return _context2.abrupt("return", this);

        case 1:
        case "end":
          return _context2.stop();
      }
    }, null, this);
  };
}
```
</details>

### DIFF
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -90,8 +90,14 @@ exports.getVisitor = ({ types: t }) => ({
       // declarations with equivalent assignment expressions.
       let vars = hoist(path);
 
-      let didRenameArguments = renameArguments(path, () => t.clone(argsId));
-      if (didRenameArguments) {
+      let context = {
+        usesThis: false,
+        usesArguments: false,
+        getArgsId: () => t.clone(argsId),
+      };
+      path.traverse(argumentsThisVisitor, context);
+
+      if (context.usesArguments) {
         vars = vars || t.variableDeclaration("var", []);
         const argumentIdentifier = t.identifier("arguments");
         // we need to do this as otherwise arguments in arrow functions gets hoisted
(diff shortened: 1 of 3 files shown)


packages/regenerator-transform/src/visit.js
```diff
@@ -108,16 +114,22 @@ exports.getVisitor = ({ types: t }) => ({
...
-        t.thisExpression()
-      ];
-
-      let tryLocsList = emitter.getTryLocsList();
+        wrapArgs.push(t.nullLiteral());
+      }
+      if (context.usesThis) {
+        wrapArgs.push(t.thisExpression());
+      } else if (tryLocsList) {
+        wrapArgs.push(t.nullLiteral());
```
benjamn: I anticipated this part might be tricky because the `this` expression is often the last argument, but not always. Clearly you figured out what was going on!

packages/regenerator-transform/src/visit.js
```diff
@@ -243,31 +255,20 @@ function getMarkedFunctionId(funPath) {
...
   Identifier: function(path, state) {
     if (path.node.name === "arguments" && util.isReference(path)) {
       util.replaceWithOrRemove(path, state.getArgsId());
-      state.didRenameArguments = true;
+      state.usesArguments = true;
     }
+  },
+
+  ThisExpression: function(path, state) {
+    state.usesThis = true;
```
benjamn: I suppose we could abandon the traversal once both `state.usesArguments` and `state.usesThis` have been set to `true`, though I'm not sure that case will be common enough to justify the extra complexity.

packages/regenerator-transform/src/visit.js
```diff
@@ -312,3 +313,13 @@ let awaitVisitor = {
     ));
   }
 };
+
+let thisVisitor = {
```
benjamn: Where is this visitor used, now?

packages/regenerator-transform/src/visit.js
nicolo-ribaudo: We can't, because `util.replaceWithOrRemove(path, state.getArgsId());` needs to be run even if `usesArguments` is already `true`.

packages/regenerator-transform/src/visit.js
nicolo-ribaudo: Nowhere :+1:

PR #369, opened 2019-03-07
Define regeneratorRuntime globally as a side-effect, again.

### LINKED ISSUE #337: `regenerator-runtime/runtime` after `regenerator-runtime` does not work
`require('regenerator-runtime/runtime')` does not work if `regenerator-runtime` has been previously required.

```js
require('regenerator-runtime')

// regeneratorRuntime is undefined
console.log(global.regeneratorRuntime)

require('regenerator-runtime/runtime')

// regeneratorRuntime is still undefined
console.log(global.regeneratorRuntime)
```

[Repro repo](https://github.com/julien-f/regenerator-runtime-issue).

### LINKED ISSUE #363: Global variable is not defined in babel 6 / webpack 3
I'm using babel v6.26.3 and webpack 3.12.0

If I call regenerrator-runtime as `import 'regenerator-runtime/runtime'` to define a global variable the code results to be
START
```javascript
/***/ "./node_modules/regenerator-runtime/runtime.js":
/***/ (function(module, exports, __webpack_require__) {

/**
 * Copyright (c) 2014-present, Facebook, Inc.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */

var regeneratorRuntime = (function (exports) {
  "use strict";

  var Op = Object.prototype;
```

END
```javascript
  // Regardless of whether this script is executing as a CommonJS module
  // or not, return the runtime object so that we can declare the variable
  // regeneratorRuntime in the outer scope, which allows this module to be
  // injected easily by `bin/regenerator --include-runtime script.js`.
  return exports;

}(
  // If this script is executing as a CommonJS module, use module.exports
  // as the regeneratorRuntime namespace. Otherwise create a new empty
  // object. Either way, the resulting object will be used to initialize
  // the regeneratorRuntime variable at the top of this file.
   true ? module.exports : {}
));


/***/ }),
```

Which means "regeneratorRuntime" is never registered as a global variable and the code stops to work.

This is happening in 0.13.1, I moved to 0.12.1 and everything is working fine there.

### LINKED ISSUE #367: improve doc for runtime
import 'regenerator-runtime/runtime' or require('regenerator-runtime/runtime') won't define a global regeneratorRuntime after 0.13

fix it or document it, please

benjamn:
In #353, I simplified the implementation of `regenerator-runtime` by converting `runtime.js` to a normal CommonJS module. No more global object detection, and no more sneaky capture/deletion of `regeneratorRuntime`.

Unfortunately, folks who use only `@babel/polyfill` and not `@babel/transform-runtime` may have existing code that relies on a global `regeneratorRuntime` variable. That code will eventually disappear (🤞), but in the meantime we can continue supporting it by simply ensuring that `regeneratorRuntime` is defined globally as a side-effect of importing `regenerator-runtime`.

This turns out to be much easier than detecting the global object, as long as `runtime.js` runs as a non-strict CommonJS module. See my code comments for another fallback solution that should cover all remaining cases.

Background discussion: https://github.com/babel/babel/pull/7646#issuecomment-468913977

Fixes #363.
Fixes #337.
Fixes #367.

### DIFF
--- a/packages/regenerator-runtime/package.json
+++ b/packages/regenerator-runtime/package.json
@@ -10,6 +10,7 @@
     "generator",
     "async"
   ],
+  "sideEffects": true,
   "repository": {
     "type": "git",
     "url": "https://github.com/facebook/regenerator/tree/master/packages/regenerator-runtime"
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-var regeneratorRuntime = (function (exports) {
+var runtime = (function (exports) {
   "use strict";
 
   var Op = Object.prototype;
@@ -709,3 +709,18 @@ var regeneratorRuntime = (function (exports) {
   // the regeneratorRuntime variable at the top of this file.
   typeof module === "object" ? module.exports : {}
 ));
+
+try {
+  regeneratorRuntime = runtime;
+} catch (accidentalStrictMode) {
+  // This module should not be running in strict mode, so the above
+  // assignment should always work unless something is misconfigured. Just
+  // in case runtime.js accidentally runs in strict mode, we can escape
+  // strict mode using a global Function call. This could conceivably fail
+  // if a Content Security Policy forbids using Function, but in that case
+  // the proper solution is to fix the accidental strict mode problem. If
+  // you've misconfigured your bundler to force strict mode and applied a
+  // CSP to forbid Function, and you're not willing to fix either of those
+  // problems, please detail your unique predicament in a GitHub issue.
+  Function("r", "regeneratorRuntime = r")(runtime);
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 before_install:
   - npm install --global npm@latest
 
+# Prevent Travis from running `npm ci` by default, which has trouble with
+# file: dependencies.
+install:
+  - npm install
+
 language: node_js
 node_js:
   - "10"
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,34 +13,34 @@
       }
     },
     "@babel/core": {
-      "version": "7.1.6",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.1.6.tgz",
-      "integrity": "sha512-Hz6PJT6e44iUNpAn8AoyAs6B3bl60g7MJQaI0rZEar6ECzh6+srYO1xlIdssio34mPaUtAb1y+XlkkSJzok3yw==",
+      "version": "7.3.4",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.3.4.tgz",
+      "integrity": "sha512-jRsuseXBo9pN197KnDwhhaaBzyZr2oIcLHHTt2oDdQrej5Qp57dCCJafWx5ivU8/alEYDpssYqv1MUqcxwQlrA==",
       "requires": {
         "@babel/code-frame": "^7.0.0",
-        "@babel/generator": "^7.1.6",
-        "@babel/helpers": "^7.1.5",
-        "@babel/parser": "^7.1.6",
-        "@babel/template": "^7.1.2",
-        "@babel/traverse": "^7.1.6",
-        "@babel/types": "^7.1.6",
+        "@babel/generator": "^7.3.4",
+        "@babel/helpers": "^7.2.0",
+        "@babel/parser": "^7.3.4",
+        "@babel/template": "^7.2.2",
+        "@babel/traverse": "^7.3.4",
+        "@babel/types": "^7.3.4",
         "convert-source-map": "^1.1.0",
         "debug": "^4.1.0",
         "json5": "^2.1.0",
-        "lodash": "^4.17.10",
+        "lodash": "^4.17.11",
         "resolve": "^1.3.2",
         "semver": "^5.4.1",
         "source-map": "^0.5.0"
       }
     },
     "@babel/generator": {
-      "version": "7.1.6",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.1.6.tgz",
-      "integrity": "sha512-brwPBtVvdYdGxtenbQgfCdDPmtkmUBZPjUoK5SXJEBuHaA5BCubh9ly65fzXz7R6o5rA76Rs22ES8Z+HCc0YIQ==",
+      "version": "7.3.4",
+      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.3.4.tgz",
+      "integrity": "sha512-8EXhHRFqlVVWXPezBW5keTiQi/rJMQTg/Y9uVCEZ0CAF3PKtCCaVRnp64Ii1ujhkoDhhF1fVsImoN4yJ2uz4Wg==",
       "requires": {
-        "@babel/types": "^7.1.6",
+        "@babel/types": "^7.3.4",
         "jsesc": "^2.5.1",
-        "lodash": "^4.17.10",
+        "lodash": "^4.17.11",
         "source-map": "^0.5.0",
         "trim-right": "^1.0.1"
       }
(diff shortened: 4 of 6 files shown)


packages/regenerator-runtime/package.json
```diff
@@ -10,6 +10,7 @@
     "generator",
     "async"
   ],
+  "sideEffects": true,
```
benjamn: This is to keep Webpack from removing `import … from "regenerator-runtime"`, since the package now has an important side-effect.

packages/regenerator-runtime/runtime.js
```diff
@@ -709,3 +709,18 @@ var regeneratorRuntime = (function (exports) {
...
+} catch (accidentalStrictMode) {
+  // This module should not be running in strict mode, so the above
+  // assignment should always work unless something is misconfigured. Just
+  // in case runtime.js accidentally runs in strict mode, we can escape
+  // strict mode using a global Function call. This could conceivably fail
+  // if a Content Security Policy forbids using Function, but in that case
+  // the proper solution is to fix the accidental strict mode problem. If
+  // you've misconfigured your bundler to force strict mode and applied a
+  // CSP to forbid Function, and you're not willing to fix either of those
+  // problems, please detail your unique predicament in a GitHub issue.
```
nicolo-ribaudo: CSP errors are normal errors, right? So people which see it will follow the first entry in the stack trace and find this comment?

benjamn: That’s right, they’re runtime errors due to actually creating and calling a `Function`, not some sort of static error based on the appearance of `Function` in the code, as far as I know.

zloirock:
I can't understand, why you against simple global object detection like [this](https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/global.js)? With this code, you will have fewer issues than with code from this PR - now someone definitely will try to load it in strict mode.

benjamn:
You're throwing around words like "simple" and "definitely" based on your intuition, and I think you reasoning may be biased because you have your own implementation.

I prefer this approach because it doesn't involve obtaining a reference to the global object. I've been down that road before (e.g. #336), and I don't care to go there again.

All the identifiers you're checking in `global.js` could be shadowed by malicious code, and you're falling back to `Function("return this")()` for Node.js, which doesn't have `window` or `self`.

If someone happens to load this code in strict mode, that's fine. It will fall back to `Function("r", "regeneratorRuntime = r")(runtime)`, which escapes strict mode. I don't even think that fallback is necessary, since loading this code in strict mode is a bug, but this implementation covers that case too, just to be safe.

If you're wondering why I didn't just depend on `core-js` here… this code has no other dependencies, which is important because it gets used in non-module contexts, such as the command-line [`regenerator`](https://github.com/facebook/regenerator/blob/master/bin/regenerator) script.

zloirock:
> You're throwing around words like "simple" and "definitely" based on your intuition, and I think you reasoning may be biased because you have your own implementation.

It's based on the experience of support modules which need the same functionality and many issues in the `babel` repo related strict mode. A long time ago `core-js` had more issues with the global object reference with many interesting environments than `regenerator` with #336, but the proposed code does not cause any issues more than 3 years.

> It will fall back to `Function...`, which escapes strict mode.

And will cause a CSP issue in strict mode.

> If you're wondering why I didn't just depend on `core-js` here…

I understand it and I do not insist - `core-js` also hasn't any dependencies for many reasons.

This PR is a solution to my issue, but I told about global object since just wanted to prevent some future issues.

MattiasBuelens:
With this change, `@babel/plugin-transform-runtime` would also pollute the global scope with a `regeneratorRuntime` variable, which kind of defeats the purpose of that plugin. 😕 

I think we should go back to having two versions: `runtime.js` (for browsers) and `runtime-module.js` (for CommonJS). The problem with this original approach was `runtime-module.js` needed to restore the old version of `regeneratorRuntime` after it had `require`'d `runtime.js` ([source](https://github.com/facebook/regenerator/blob/fde052c6c3563e007d233870ecd4a41920773bbd/packages/regenerator-runtime/runtime-module.js)).

What if we turned it around?
* `runtime-module.js` contains the actual implementation. It defines a local variable `runtime` and (if available) exports it to `module.exports`. This is basically [lines 1 to 711 from `runtime.js`](https://github.com/facebook/regenerator/blob/9c90408393bc8e4efc3b715d9e08a80c5f8367b9/packages/regenerator-runtime/runtime.js#L1-L711).
* `runtime.js` is generated by a script that concatenates `runtime-module.js` together with an "outro" that defines the global `regeneratorRuntime` variable ([lines 713 to 726](https://github.com/facebook/regenerator/blob/9c90408393bc8e4efc3b715d9e08a80c5f8367b9/packages/regenerator-runtime/runtime.js#L713-L726)). (We can't `require("./runtime-module.js")` since that wouldn't work inside a browser.) The result would be identical to the current version of `runtime.js`.

Usage:
* CommonJS users would look at the `main` field of `package.json` to load `runtime-module.js`.
* Browser users would need to `require("regenerator-runtime/runtime")`, as they did before version 0.13.0. (Alternatively, we could add a `browser` field in `package.json` for module loaders that support it.)

OurMajesty:
For some reason regenerator-runtime@latest is 0.13.1
изображение

benjamn:
@OurMajesty That's right, I didn't want people updating to this version accidentally before we were sure the changes worked.

@MattiasBuelens With that solution, I would be worried that people would end up bundling both `runtime-module.js` and `runtime.js`, thus duplicating the code. Compared to that very real risk, I don't see the "pollution" of global scope with an additional variable as much of a problem. The purpose of the runtime transform, in my mind, is not so much to avoid polluting global scope, but to ensure you don't have to import/bundle `regenerator-runtime` if you don't use any generator functions (the "pay for what you use" principle). That property still holds with (or without) this change, for applications that use `@babel/plugin-transform-runtime`. This change simply makes `@babel/polyfill` work the way it previously did, for applications that still use that approach. The global variable is a small price to pay for backwards compatibility.

MattiasBuelens:
> The purpose of the runtime transform, in my mind, is not so much to avoid polluting global scope, but to ensure you don't have to import/bundle `regenerator-runtime` if you don't use any generator functions (the "pay for what you use" principle).

Well, the documentation for `@babel/plugin-transform-runtime` recommends it for library development [because it should not pollute the global scope](https://babeljs.io/docs/en/babel-plugin-transform-runtime#why):
> Another purpose of this transformer is to create a sandboxed environment for your code. If you use `@babel/polyfill` and the built-ins it provides such as `Promise`, `Set` and `Map`, those will pollute the global scope. While this might be ok for an app or a command line tool, it becomes a problem if your code is a library which you intend to publish for others to use or if you can't exactly control the environment in which your code will run.

Sure, an extra global variable doesn't matter when you're bundling a web app. However, if you're bundling a web _library_, and users use your library inside their own web app that uses `@babel/polyfill`, it could mean that users accidentally use _your_ version of the runtime instead of the one from their polyfill.

Of course, in this case there's also the problem of bundling the runtime twice (once in the library, once in the app), so I guess library developers should recommend their users to bundle the library's dependencies themselves...? 😕 

Either way, I understand you don't want to risk bundling the runtime twice. I think it's unfortunate for library developers, but I guess that's the trade-off we have to make.

zloirock:
@benjamn could you publish a release with those changes as `latest`? Without it, we will have problems since `babel@7.4` which requires it already released.

nicolo-ribaudo:
Well, Babel downloads the correct version anyway because it uses `^0.13.2`

zloirock:
@nicolo-ribaudo I mean the case of separate usage of `core-js` and `regenerator-runtime` from the `latest` branch in `preset-env`.

benjamn:
I just set 0.13.2 to be `latest` with the following command:
```sh
npm dist-tag add regenerator-runtime@0.13.2 latest
```
The `next` tag still points to that version, too, in case anyone was relying on that.

packages/regenerator-runtime/runtime.js
bigmeow: I think should handle exception for this line code
```js
try {
  Function("r", "regeneratorRuntime = r")(runtime);
} catch (error) {
  console.warn('something')
}

```

PR #370, opened 2019-03-21
docs: Update runtime README

taion:
There's no separate module export any more, so there's no point in explicitly importing the runtime path.


taion:
Ping!

Issue #354, opened 2018-08-24
Package distributions are not licensed


honzajavorek:
Hi @benjamn 👋 The code of the `regenerator` monorepo is licensed under MIT, and that's great. But according to MIT, the license text needs to be attached everywhere:

> The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.

The npm packages as they're published and distributed, do not comply with this as they do not contain the license files. Effectively, without the full license text they're proprietary code and cannot be used by anyone who cares about licenses. The SPDX identifier in the `package.json` is not satisfactory (not only) for the reasons mentioned above. There are two solutions to this:

1. Upgrade to Lerna@3, as it has the licensing built-in now
2. Copy & paste the root license to all projects in the `packages` directory, so it gets picked up by npm during publishing, and to re-publish all of them with a new patch version.

We did the second with Babel as it was faster than to upgrade Lerna, atm. Especially in this particular case, when there are just 3 packages, I think just copypasting the license to those would be a very fast and simple solution.

For more information, see https://github.com/lerna/lerna/pull/1465#issuecomment-405184876, https://github.com/babel/babel/pull/7308#issuecomment-367615152, https://github.com/babel/babel/pull/8409#issuecomment-415379558.

-----

Side note: I found some related issues, which I believe are obsolete and I think they can be closed right away: https://github.com/facebook/regenerator/pull/215 by @rafaelstz, https://github.com/facebook/regenerator/issues/298 by @krotscheck

Bosch-Eli-Black:
Any chance of getting this fixed? It looks like this is still happening in the latest versions.

hrieke:
Upvote.
Com'on don't be a lazy programmer...

PR #376, opened 2019-07-02
Appeasing corporate legal department

DiscoNova:
Corporate legal was unsatisfied with the license only being found package.json and not directly accessible in full from within the individual distributed packages (since "the root directory of this source tree" is not present in the distributed packages). This pull request adds the license-files to them as well as updates the patch-levels of the distributed package versions in preparation for `npm publish`.

Resolves: #354

### DIFF
--- a/packages/regenerator-preset/LICENSE
+++ b/packages/regenerator-preset/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2014-present, Facebook, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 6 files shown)


DiscoNova:
> I know symbolic links can be troublesome on Windows

Yep, I had the same concern in mind and decided to do a plain and simple copy of the files.

However, unless I'm mistaken (which is always a possibility), Git does only store files with matching content only once saving bandwidth on checkout, but ... of course, this doesn't help developers who will still end up with multiple copies of the same file on their hard drives (which may eventually lead to said files diverging if one file gets changes while others might not).

PR #377, opened 2019-07-15
Register new bindings when hoisting vars

nicolo-ribaudo:
In https://github.com/facebook/regenerator/pull/368, I changed `regenerator-transform` to remove bindings which are hoisted, since it caused problems when they were registered again.
Nothing in this plugin actually registered them, but sometimes Babel re-traverses the scope and registers declaration which weren't already registerd (this usually happens when a function node is discarded and re-created, I think). For this reason, this code used to throw:
```js
async function foo() {
	(async function (number) {
		const tmp = number
	})
}
```
because when Babel did re-crawl the bindings, it already had a `tmp` binding registered (for `const tmp`) and tried to register a new one with the same name (for the hoisted `var tmp`).

Unfortunately we can't always rely on something to trigger the scope update, so simply deleting the old binding isn't enough. We also need to ensure that the new hoisted `var`s get registered again. Otherwise, we cause conflicts with the `transform-modules-commonjs` plugin (or any other plugin relying on scope information):

```js
import { someAction } from 'actions';

function* foo() {
  const someAction = bar;
}
```

This code is initially transformed by regenerator to something like this:

```js
import { someAction } from 'actions';

function foo() {
  var someAction;

  return regeneratorRuntime.wrap(function foo() {
    someAction = bar;
  });
}
```

it correctly removes the `const someAction` binding from the scope, but before this PR it didn't register the `var someAction` binding. For this reason, the modules transform plugin thought that `someAction = bar` was assigning to the imported binding (which is read-only), generating an error.

Fixes babel/babel#10193

---

PS: Do you have a custom `.npmrc` for this project on your machine? Every time I run `npm test` it generates `package-lock.json` files for the subpackages, and it changes the `dependencies` section in the top-level `package.json` file to reference the local packages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -114,6 +114,27 @@
         "@babel/types": "^7.0.0"
       }
     },
+    "@babel/helper-module-imports": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/helper-module-imports/-/helper-module-imports-7.0.0.tgz",
+      "integrity": "sha512-aP/hlLq01DWNEiDg4Jn23i+CXxW/owM4WpDLFUbpjxe4NS3BhLVZQ5i7E0ZrxuQ/vwekIeciyamgB1UIYxxM6A==",
+      "requires": {
+        "@babel/types": "^7.0.0"
+      }
+    },
+    "@babel/helper-module-transforms": {
+      "version": "7.4.4",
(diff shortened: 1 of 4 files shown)


pwnreza:
Testet it. In case of [babel/10193](https://github.com/babel/babel/issues/10193) it works like a charm.
Big thanks to @nicolo-ribaudo

PR #382, opened 2019-12-11
add `.js` extension to file import

JLHwung:
Importing files in a package requires explicit file extension unless it is defined as package exports, see https://nodejs.org/api/esm.html#esm_package_scope_and_file_extensions


nicolo-ribaudo:
People using native modules don't need to use regenerator, so the docs are only for who already transpiles ESM.
Also, I don't think that regenerator works inside native ES modules 🤔

benjamn:
> I don't think that regenerator works inside native ES modules

Why would that be, besides it not being necessary (because all JS environments with ESM support also have generator support)?

PR #383, opened 2020-01-13
Pass Promise as a parameter to regeneratorRuntime.async()

nicolo-ribaudo:
This makes it possible to run `@babel/plugin-transform-runtime` and `@babel/preset-env`'s `useBuiltIns` on `regenerator`'s output to provide a Promise polyfill in old browsers. 

Fixes babel/babel#10678, babel/babel#10830

### DIFF
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -207,7 +207,7 @@ var runtime = (function (exports) {
   // AsyncIterator objects; they just return a Promise for the value of
   // the final result produced by the iterator.
   exports.async = function(innerFn, outerFn, self, tryLocsList, PromiseImpl) {
-    if (PromiseImpl === undefined) PromiseImpl = Promise;
+    if (PromiseImpl === void 0) PromiseImpl = Promise;
 
     var iter = new AsyncIterator(
       wrap(innerFn, outerFn, self, tryLocsList),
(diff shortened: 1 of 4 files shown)


packages/regenerator-runtime/runtime.js
```diff
@@ -206,9 +206,12 @@ var runtime = (function (exports) {
   // Note that simple async functions are implemented on top of
   // AsyncIterator objects; they just return a Promise for the value of
   // the final result produced by the iterator.
-  exports.async = function(innerFn, outerFn, self, tryLocsList) {
+  exports.async = function(innerFn, outerFn, self, tryLocsList, PromiseImpl) {
+    if (PromiseImpl === undefined) PromiseImpl = Promise;
```
benjamn:
```suggestion
    if (PromiseImpl === void 0) PromiseImpl = Promise;
```
A little shorter, if you like.

benjamn:
@nicolo-ribaudo Do you think this should be a minor version bump for `regenerator-runtime` and/or `regenerator-transform`, or is it fully backwards compatible?

nicolo-ribaudo:
This is fully backward compatible.

It could only break if someone did already pass a fifth parameter to `regeneratorRuntime.async`, but since no one calls it directly (it's only called by the generated code), it shouldn't be a problem in practice.

nicolo-ribaudo:
Thank you!

PR #385, opened 2020-03-05
Add array spread support

nicolo-ribaudo:
There are some browsers that support array spread but don't support generators.

**Why did we never realize it?**

The browsers versions that support array spread and the versions that supports generators are _close_:

```json
"transform-spread": {
  "chrome": "46",
  "edge": "13",
  "firefox": "36",
  "safari": "10",
  "node": "5",
  "ios": "10",
  "samsung": "5",
  "opera": "33",
  "electron": "0.36"
},
"transform-regenerator": {
  "chrome": "50",
  "edge": "13",
  "firefox": "53",
  "safari": "10",
  "node": "6",
  "ios": "10",
  "samsung": "5",
  "opera": "37",
  "electron": "1.1"
},
```

Probably no one ever used a browsers query in `preset-env` that included `transform-regenerator` but not `transform-spread`.

**Why did we found this bug now?**

`create-react-app` defaults to `>0.2%, not dead, not op_mini all` for browsers support. Some users additionally exclude IE 11 and Safari 5.1, since they are really old and force the inclusion of a lot of different transforms.
Due to some recent changes in the usage data on caniuse/browserslist (an old browser felt below 0.2%, or it become "dead"), this currently matches [these browsers](https://browsersl.ist/?q=%3E0.2%25):
- Chrome 49
- Edge 17
- Firefox 70
- Opera 63
- Safari 11.1

These browsers need `transform-regenerator` (because of Chrome 49), but they don't need `transform-spread`.

**Alternative solution**

We could force `transform-spread` when using `transform-regenerator` in `preset-env`, but I'd avoid forcing an unnecessary transform if possible.

---

Fixes https://github.com/babel/babel/issues/11212

### DIFF
--- a/test/run.js
+++ b/test/run.js
@@ -165,11 +165,10 @@ enqueue(convert, [
   "./test/class.es5.js"
 ]);
 
-function convertWithSpread(es6File, es5File, callback) {
+function convertWithParamsTransform(es6File, es5File, callback) {
   var transformOptions = {
     presets:[require("regenerator-preset")],
     plugins: [
-      require("@babel/plugin-transform-spread"),
       require("@babel/plugin-transform-parameters")
     ],
     parserOpts: {
@@ -200,7 +199,7 @@ function convertWithSpread(es6File, es5File, callback) {
   });
 }
 
-enqueue(convertWithSpread, [
+enqueue(convertWithParamsTransform, [
   "./test/regression.js",
   "./test/regression.es5.js"
 ]);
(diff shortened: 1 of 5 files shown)


test/run.js
```diff
@@ -165,11 +165,10 @@ enqueue(convert, [
   "./test/class.es5.js"
 ]);
 
-function convertWithSpread(es6File, es5File, callback) {
+function convertWithParamsTransform(es6File, es5File, callback) {
```
benjamn: I really appreciate your attention to detail, such as taking the time to update the test harness so it doesn't accidentally rely on the `@babel/plugin-transform-spread` transform.

PR #388, opened 2020-03-17
Remove _shadowFunctionLiteral use

### LINKED ISSUE #387: Consider removing _shadowedFunctionLiteral use in regenerator-transform
Located at [0]. Attaching a `NodePath` to a node causes the Babel ast not to be serializable with `v8.serialize` (it contains at least one function) nor with `JSON.stringify` (it makes the structure circular). We'd like to serialize and persist Babel asts in parcel-bundler/parcel, but this transform prevents us from doing that.

While it looks like at one point this had special meaning to babel [1], it looks like it doesn't anymore [2]. Removing this line doesn't cause the test added in #283 to fail. Is it still needed?

cc @devongovett @padmaia

[0] https://github.com/facebook/regenerator/blob/11e270afaf9718c06a41493463de0ea0b70ef4e4/packages/regenerator-transform/src/visit.js#L119
[1] https://github.com/facebook/regenerator/pull/283#pullrequestreview-27500046
[2] https://github.com/babel/babel/search?q=_shadowedFunctionLiteral&unscoped_q=_shadowedFunctionLiteral

wbinnssmith:
Closes #387. Additional context is available there.

This removes the use of attaching a `_shadowFunctionLiteral` property with a `NodePath` value to a `Node`, which prevents Babel ASTs from being serialized. It doesn't appear to be used by Babel any longer and tests still pass.

Test Plan: `npm test`. Tests enforcing the Babel ast can be serialized were added. Existing tests continue to pass.

### DIFF
--- a/packages/regenerator-transform/src/visit.js
+++ b/packages/regenerator-transform/src/visit.js
@@ -114,11 +114,9 @@ exports.getVisitor = ({ types: t }) => ({
 
       if (context.usesArguments) {
         vars = vars || t.variableDeclaration("var", []);
-        const argumentIdentifier = t.identifier("arguments");
-        // we need to do this as otherwise arguments in arrow functions gets hoisted
-        argumentIdentifier._shadowedFunctionLiteral = path;
         vars.declarations.push(t.variableDeclarator(
-          t.clone(argsId), argumentIdentifier
+          t.clone(argsId),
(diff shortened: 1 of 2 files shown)


wbinnssmith:
cc @benjamn @danez @devongovett @padmaia

PR #390, opened 2020-04-08
polish: print friendly error when hoisting unsupported node types

JLHwung:
Context: https://github.com/babel/babel/issues/10808

Transform the following snippet with only `regenerator-transform`
```js
async (req) => {
  const { params } = req;
}
```
 will throw
```
TypeError: /src.js: Property name expected type of string but got null
    at validate (/node_modules/@babel/types/lib/definitions/utils.js:164:13)
    at Object.validate (/node_modules/@babel/types/lib/definitions/utils.js:201:7)
    at validateField (node_modules/@babel/types/lib/validators/validate.js:22:9)
    at validate (/node_modules/@babel/types/lib/validators/validate.js:16:3)
    at builder (/node_modules/@babel/types/lib/builders/builder.js:38:27)
    at Object.Identifier (/node_modules/@babel/types/lib/builders/generated/index.js:334:31)
    at /node_modules/regenerator-transform/lib/hoist.js:32:29
```

Since practically every engine with native destructuring support will also have native generator support ([thread](https://github.com/babel/babel/issues/10808#issuecomment-611119555)). I think the value of pattern support is marginal. Hence we throw a better error.

### DIFF
--- a/packages/regenerator-transform/src/hoist.js
+++ b/packages/regenerator-transform/src/hoist.js
@@ -24,6 +24,10 @@ exports.hoist = function(funPath) {
     let exprs = [];
 
     vdec.declarations.forEach(function(dec) {
+      if (!t.isIdentifier(dec.id)) {
+        // We don't support ArrayPattern and ObjectPattern
+        throw new Error("Unsupported declarator with type " + JSON.stringify(dec.id.type) + ", please add \"@babel/plugin-transform-destructuring\" to your babel config.")
+      }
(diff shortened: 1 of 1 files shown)


monsonjeremy:
🙌

tenjojeremy:
How is this error fixed?

I get the error with the following code

```js
const hello = async () => {
  const he = { ll: '2' }
  const { ll } = he
}
```

Thanks

monsonjeremy:
Any chance we can get this merged?

PR #394, opened 2020-04-23
Relieve dependency on private package

kriskowal:
Modern JS engines now support WeakMap.  WeakMaps solve the same problem as the private package, but without needing to tamper with intrinsics to ensure their privacy.

Relaxing this dependency eases compatibility with SES (a secure ECMAScript subset that leaves intrinsics immutable).

### DIFF
--- a/packages/regenerator-transform/package.json
+++ b/packages/regenerator-transform/package.json
@@ -32,8 +32,7 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


kriskowal:
Ahoy @benjamn. Please let me know if there’s anything I can do to help release this change. This limits the use of Babel in a secured JavaScript environment.

PR #399, opened 2020-07-21
runtime: Use `[[Define]]` for the `Symbol.toStringTag` property

ExE-Boss:
See <https://github.com/tc39/proposal-iterator-helpers/issues/115> and **Firefox**’s [Bug 1644581 (comment #6)](https://bugzilla.mozilla.org/show_bug.cgi?id=1644581#c6).

### DIFF
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -8,6 +8,21 @@
 var runtime = (function (exports) {
   "use strict";
 
+  // Inlined from `@babel/runtime/helpers/defineProperty`
+  function _defineProperty(obj, key, value) {
+    if (key in obj) {
+      Object.defineProperty(obj, key, {
+        value: value,
+        enumerable: true,
+        configurable: true,
+        writable: true
+      });
+    } else {
+      obj[key] = value;
+    }
+    return obj;
+  }
+
   var Op = Object.prototype;
   var hasOwn = Op.hasOwnProperty;
   var undefined; // More compressible than void 0.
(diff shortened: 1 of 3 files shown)


ljharb:
I filed #400 which uses a conditional; defineProperty won't work in IE < 9, which is an important target for this package.

ExE-Boss:
This will work in **IE &lt; 9**, since it only uses `defineProperty` if the symbol property already exists in `GeneratorPrototype`’s prototype chain.

benjamn:
That's a fair point @ExE-Boss! Do you see value in overriding the existing string, if it's already defined? PR #400 lets the existing string remain unchanged, whereas `defineProperty` would presumably change it. I have no strong opinion here.

ExE-Boss:
Well, PR <https://github.com/facebook/regenerator/pull/400> makes it so that once browsers start shipping `Iterator.prototype[Symbol.toStringTag]`, then `Object.prototype.toString` will return `[object Iterator]` instead of `[object Generator]` for regenerator’s generator objects.

ljharb:
That sounds like the right behavior to me, since that's what the standard will dictate.

ExE-Boss:
But the `%Symbol.toStringTag%` property of `%GeneratorFunction.prototype.prototype%` is `"Generator"`.

ljharb:
I'm a bit confused.

This change was motivated by a web compat concern, where `Gp[Symbol.toStringTag]` was being assigned, but this was breaking it. Are you saying that it's because it should be defining an own property instead of using [[Set]] and hitting a non-writable property up the [[Prototype]] chain?

ExE-Boss:
> Are you saying that it's because it should be defining an own property instead of using [[Set]] and hitting a non-writable property up the [[Prototype]] chain?

Yes.

ljharb:
In that case, i think the helper @benjamn added should be modified to *conditionally* use `Object.defineProperty`, when it's available. It will need to be used in a try/catch, since IE 8 has the method, but it throws on things that aren't HTML elements.

ExE-Boss:
That’s what I’m doing in <https://github.com/facebook/regenerator/pull/402>.

PR #402, opened 2020-07-22
runtime: Use `[[Define]]` for the `Symbol.toStringTag` property

ExE-Boss:
PR <https://github.com/facebook/regenerator/pull/400> makes it so that once engines start shipping `Iterator.prototype[Symbol.toStringTag]`, then `Object.prototype.toString` will return `[object Iterator]` instead of `[object Generator]` for regenerator’s generator objects.

This fixes that to ensure that it doesn’t happen.

### DIFF
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -8,6 +8,7 @@
 var runtime = (function (exports) {
   "use strict";
 
+  var define = Object.defineProperty;
   var Op = Object.prototype;
   var hasOwn = Op.hasOwnProperty;
   var undefined; // More compressible than void 0.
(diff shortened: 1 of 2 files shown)


packages/regenerator-runtime/runtime.js
```diff
@@ -85,7 +85,13 @@ var runtime = (function (exports) {
   function ensureDefaultToStringTag(object, defaultValue) {
     // https://bugzilla.mozilla.org/show_bug.cgi?id=1644581#c6
     return toStringTagSymbol in object
-      ? object[toStringTagSymbol]
+      ? (Object.defineProperty(object, toStringTagSymbol, {
```
benjamn, line 88: Should the condition maybe just be whether `Object.defineProperty` exists, since we're setting the tag either way, regardless of whether `toStringTagSymbol in object`?

benjamn:
Can you open that repo/branch for me to push? Also happy to merge and then make my own tweaks.

ExE-Boss:
@benjamn I had sent you an invitation to @EB-Forks.

benjamn:
@ExE-Boss Ahh I missed that. Pushed (above)!

Issue #407, opened 2021-02-01
Support generators containing classes


nicolo-ribaudo:
Some browsers (e.g. firefox 50) support classes but not generators.
It would be good it `regenerator` supported untranspiled classes.

[REPL](https://babeljs.io/repl#?browsers=firefox%2050&build=&builtIns=false&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHgFRwAUAlHAN4BQccUwM6UecBmqAzvzigYwPAiEAhAcAbMWcZBAi05ChWw5c4ARgDc8lgF95Rg0aA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.12.12&externalPlugins=)

---

This issue was originally reported by @boris-petrov at https://github.com/babel/babel/issues/12692:

> ## Bug Report
> 
> **Current behavior**
> - [REPL](https://babeljs.io/repl#?browsers=firefox%2047&build=&builtIns=false&spec=false&loose=true&code_lz=KYDwDg9gTgLgBAE2AMwIYFcA29UGcCeAdgMZzLokwCWEhcAFAJRwDeAUHHFMDOlHcUx5ccUDGCEEIgEJ5grDpzIQITBUqXde_OAEYA3Is4BfRccOmgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cstage-1&prettier=true&targets=Node-14.14&version=7.12.12&externalPlugins=)
> 
> ```js
> export default async function () {
>   return class extends Base {
>     foo() {
>       return 1;
>     }
>   };
> }
> ```
> 
> **Expected behavior**
> Code to be transpiled correctly.
> 
> Currently Babel blows up with:
> 
> ```
> unknown Expression of type "ClassExpression"
> ```
> 
> Using `firefox 35` for targets works. As does `firefox 53` and above. Similarly to https://github.com/babel/babel/issues/12690.

PR #408, opened 2021-02-19
Fix class support inside generators

nicolo-ribaudo:
Classes are more supported than generators, so it's possible that Babel users only compile generators:

```json
  "transform-regenerator": {
    "chrome": "50",
    "opera": "37",
    "edge": "13",
    "firefox": "53",
    "safari": "10",
    "node": "6",
    "ios": "10",
    "samsung": "5",
    "electron": "1.1"
  },
  "transform-classes": {
    "chrome": "46",
    "opera": "33",
    "edge": "13",
    "firefox": "45",
    "safari": "10",
    "node": "5",
    "ios": "10",
    "samsung": "5",
    "electron": "0.36"
  },
```

Compiling generators inside uncompiled classes is already supported (generator methods), this PR adds support for uncompiled classes inside generators.

Fixes https://github.com/facebook/regenerator/issues/407

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ test/mocha.js
 test/mocha.css
 
 test/*.es5.js
+test/*.regenerator.js
 test/tests.browser.js
 
 .idea
--- a/packages/transform/src/emit.js
+++ b/packages/transform/src/emit.js
@@ -764,6 +764,10 @@ Ep.explodeStatement = function(path, labelId) {
 
     break;
 
+  case "ClassDeclaration":
+    self.emit(self.explodeClass(path));
+    break;
+
   default:
     throw new Error(
       "unknown Statement of type " +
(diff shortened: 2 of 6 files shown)


packages/regenerator-transform/src/emit.js
```diff
@@ -900,6 +904,50 @@ Ep.updateContextPrevLoc = function(loc) {
...
+// In order to save the rest of explodeExpression from a combinatorial
+// trainwreck of special cases, explodeViaTempVar is responsible for
+// deciding when a subexpression needs to be "exploded," which is my
+// very technical term for emitting the subexpression as an assignment
+// to a temporary variable and the substituting the temporary variable
+// for the original subexpression. Think of exploded view diagrams, not
+// Michael Bay movies. The point of exploding subexpressions is to
+// control the precise order in which the generated code realizes the
+// side effects of those subexpressions.
+Ep.explodeViaTempVar = function(tempVar, childPath, hasLeapingChildren, ignoreChildResult) {
```
nicolo-ribaudo, line 917: This code is just extracted from `explodeExpression`.

JLHwung (approved):

benjamn:
Thanks for diving into this, and sorry for the long wait @nicolo-ribaudo!

While attempting to add a few more tests, I think I discovered a bug in `@babel/plugin-transform-classes`: https://github.com/facebook/regenerator/pull/487

Since this seems to be an upstream bug, I don't think it needs to hold up the release of these changes.

benjamn:
Just published these changes to npm as `regenerator-transform@0.15.0`, though I used the `next` tag for now instead of `latest`.

PR #411, opened 2021-03-22
Runtime - Support frozen intrinsics

### LINKED ISSUE #410: runtime - class inheritance pattern not compatible with frozen intrinsics
it is becoming increasingly common to freeze javascript intrinsics (Object, Array...) (see [nodejs](https://nodejs.org/api/cli.html#cli_frozen_intrinsics), [endo/ses](https://github.com/endojs/endo/tree/master/packages/ses), [xs](https://blog.moddable.com/blog/secureprivate/))

while this is generally not an issue for packages that do not extend intrinsics, there are a few unfortunate side effects due to javascript's ["override mistake"](https://github.com/tc39/ecma262/pull/1307), namely `obj.constructor = ...` fails due to the inherited property "constructor" being read-only on `Object.prototype`

here is the inheritance pattern at fault
https://github.com/facebook/regenerator/blob/0c2aba1af78be03da05de96b6c69f231b85993dc/packages/regenerator-runtime/runtime.js#L105-L106

here is a compatible inheritance pattern
https://github.com/isaacs/inherits/blob/9a2c29400c6d491e0b7beefe0c32efa3b462545d/inherits_browser.js#L6-L13
alternatively you can just use defineProperty instead of assignment

reproduction steps:
```js
Object.freeze(Object.prototype)
require('./runtime.js')
```

kumavis:
Fixes #410 

thanks for taking the time to review this change :heart: 
I wasn't sure exactly what platform to target with the tests, feedback welcome

these seemed like the least invasive change to make this work
I tested this locally against a more significant freezing of the intrinsics, built from nodejs internal [freeze_intrinsics](https://github.com/nodejs/node/blob/7c808170f5bf27b2d4a83ae5934b5e231bb5c1bc/lib/internal/freeze_intrinsics.js) and [primordials](https://github.com/nodejs/node/blob/7c808170f5bf27b2d4a83ae5934b5e231bb5c1bc/lib/internal/per_context/primordials.js)
here is a draft PR that includes those additional test utilities https://github.com/facebook/regenerator/pull/412
those tests pass with no additional change to the runtime

### DIFF
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -102,8 +102,9 @@ var runtime = (function (exports) {
 
   var Gp = GeneratorFunctionPrototype.prototype =
     Generator.prototype = Object.create(IteratorPrototype);
-  GeneratorFunction.prototype = Gp.constructor = GeneratorFunctionPrototype;
-  GeneratorFunctionPrototype.constructor = GeneratorFunction;
+  GeneratorFunction.prototype = GeneratorFunctionPrototype;
+  define(Gp, "constructor", GeneratorFunctionPrototype);
+  define(GeneratorFunctionPrototype, "constructor", GeneratorFunction);
   GeneratorFunction.displayName = define(
     GeneratorFunctionPrototype,
     toStringTagSymbol,
@@ -416,9 +417,9 @@ var runtime = (function (exports) {
     return this;
   };
 
-  Gp.toString = function() {
+  define(Gp, 'toString', function() {
     return "[object Generator]";
-  };
+  });
 
   function pushTryEntry(locs) {
     var entry = { tryLoc: locs[0] };
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -86,9 +86,9 @@ var runtime = (function (exports) {
   // This is a polyfill for %IteratorPrototype% for environments that
   // don't natively support it.
   var IteratorPrototype = {};
-  IteratorPrototype[iteratorSymbol] = function () {
+  define(IteratorPrototype, iteratorSymbol, function () {
     return this;
-  };
+  });
 
   var getProto = Object.getPrototypeOf;
   var NativeIteratorPrototype = getProto && getProto(getProto(values([])));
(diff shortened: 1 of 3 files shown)


kumavis:
cla signed

kumavis:
this needs an additional fix for this part 
```js
  Gp[iteratorSymbol] = function() {
    return this;
  };
  ```

**Comment by kumavis** [comment]
cc @kriskowal @ljharb noticed both of you in the commits and that you're familiar with the override mistake issue

File: packages/regenerator-runtime/runtime.js [hunk 1]
```diff
@@ -102,8 +102,9 @@ var runtime = (function (exports) {
 
   var Gp = GeneratorFunctionPrototype.prototype =
     Generator.prototype = Object.create(IteratorPrototype);
```
ljharb, line 104: wouldn't this line also perhaps trigger the same issue, since `.prototype` is also locked down?

ljharb (commented):
LGTM overall

packages/regenerator-runtime/runtime.js
kumavis: I believe this is covered by the added tests. function-syntax classes have their own writable `prototype` property, so it is not affected by the override mistake
```js
function Animal () {}
Reflect.getOwnPropertyDescriptor(Animal, 'prototype')
//=> Object { value: {…}, writable: true, enumerable: false, configurable: false }
```

test/run.js
```diff
@@ -141,6 +141,12 @@ if (semver.gte(process.version, "4.0.0")) {
   ]);
 }
 
+enqueue("mocha", [
+  "--harmony",
+  "--reporter", "spec",
+  "./test/frozen-intrinsics.js",
+]);
```
benjamn, line 148: @kumavis Nice, this `enqeue` will run the tests you added in an isolated child process, so the `Object.freeze` stuff doesn't leak into other tests.

benjamn (approved):
LGTM2. Thanks for tackling this @kumavis and for the supporting review @ljharb!

benjamn:
@kumavis @ljharb Just published this to npm in `regenerator-runtime@0.13.8`. However (out of an abundance of caution for a package with this many users), I used the `next` tag instead of `latest` for now. Let me know if/when you think we can promote it to `latest`. Thanks!

Issue #450, opened 2021-05-08
CSP `unsafe-eval` in Web Worker in Chrome Browser which is always strict mode


edwardxia:
https://github.com/facebook/regenerator/blob/dbbddd9bd8c099ed1cdfcecdd077b8b4e7f43042/packages/runtime/runtime.js#L736-L749

Web workers in latest Chrome stable enforces strict mode (not sure when it started to enforce), there is no way to turn it off.

Using regenerator under CSP will cause web worker to stop due to `unsafe-eval`.

Setup described in #378 are somewhat configuration issue. However, now since there is no way to turn off strict mode in Web Workers in Chrome, it is no longer a configuration issue, and need to be properly fixed.

edwardxia:
The dirty hack I currently use to workaround this issue:

```
new require('webpack').BannerPlugin({
  test: /\.worker\.js$/,
  banner: 'self.regeneratorRuntime = undefined;',
  raw: true
})
```

nicolo-ribaudo:
@benjamn Would you accept a PR to do something like this?

```js
 try { 
   regeneratorRuntime = runtime; 
 } catch (accidentalStrictMode) { 
  if (typeof globalThis === "object") {
    globalThis.regeneratorRuntime = runtime;
  } else {

   // This module should not be running in strict mode, so the above 
   // assignment should always work unless something is misconfigured. Just 
   // in case runtime.js accidentally runs in strict mode, we can escape 
   // strict mode using a global Function call. This could conceivably fail 
   // if a Content Security Policy forbids using Function, but in that case 
   // the proper solution is to fix the accidental strict mode problem. If 
   // you've misconfigured your bundler to force strict mode and applied a 
   // CSP to forbid Function, and you're not willing to fix either of those 
   // problems, please detail your unique predicament in a GitHub issue. 
   Function("r", "regeneratorRuntime = r")(runtime);
  } 
} 
```

The problem only happens in modern browsers, and they support `globalThis`.

A proper configuration solution would be to completely avoid regenerator in modern browsers, but many people have a single build and not multiple targets (and some libraries on npm rely on `regeneratorRuntime`).

benjamn:
@nicolo-ribaudo Yes, I'd be happy with that!

nicolo-ribaudo:
Awesome, I'll prepare a PR by tomorrow!

PR #480, opened 2021-07-21
[runtime] Use `globalThis` before `Function()`

nicolo-ribaudo:
Fixes https://github.com/facebook/regenerator/issues/450.


benjamn (approved):
Thanks @nicolo-ribaudo!

PR #578, opened 2022-06-23
Make `_invoke` method on generators non-enumerable

nicolo-ribaudo:
Fixes https://github.com/babel/babel/issues/14684. We cannot remove the `._invoke` method because it's needed to close over what in the spec are internal slots, but we can at least make it non-enumerable so that it's not visible to most JS APIs.

### DIFF
--- a/packages/runtime/runtime.js
+++ b/packages/runtime/runtime.js
@@ -10,6 +10,7 @@ var runtime = (function (exports) {
 
   var Op = Object.prototype;
   var hasOwn = Op.hasOwnProperty;
+  var define = Object.defineProperty || function (obj, key, desc) { obj[key] = desc.value; };
   var undefined; // More compressible than void 0.
   var $Symbol = typeof Symbol === "function" ? Symbol : {};
   var iteratorSymbol = $Symbol.iterator || "@@iterator";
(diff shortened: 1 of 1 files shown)


packages/runtime/runtime.js
```diff
@@ -10,6 +10,7 @@ var runtime = (function (exports) {
 
   var Op = Object.prototype;
   var hasOwn = Op.hasOwnProperty;
+  var define = Object.defineProperty || function (obj, key, desc) { obj[key] = desc.value; };
```
magic-akari, line 13: Check the line 20, There is already a `define` function, is this intentional?

nicolo-ribaudo: I didn't notice it, thank you!
```suggestion
```

magic-akari: But they are different. 
The `define` function at line 20 will return a enumerable method.

benjamn:
@nicolo-ribaudo This looks good to me, but it seems I no longer have the ability to approve or merge pull requests…

@zpao It looks like I missed an invitation from @facebook-github-permission-bot last month, which has now expired. Can I get invited again? 🙏 😬

benjamn (approved):
And I'm back! Thanks @zpao

nicolo-ribaudo:
I don't have write access and/or npm access, could you merge and release? :pray: 

(or @zpao, could I have write access? I have been the second top human contributor since 2017, and I'm involved in this repository because I maintain Babel)

PR #603, opened 2022-09-28
hide constructor property from prototype

JLHwung:
Fixes https://github.com/babel/babel/issues/14978. Per [spec](https://tc39.es/ecma262/#sec-generator.prototype.constructor), the `constructor` property is now marked as non-numerable and non-writable.


benjamn (approved):
Thanks @JLHwung!

PR #628, opened 2022-11-09
Fix: `yield` / `await` in sparse array

### LINKED ISSUE #627: Crashes when processing sparse array with `await`
The following code will crash `regenerator`:

```js
async () => [, await 0];
```

It will throw error (using [https://facebook.github.io/regenerator/](https://facebook.github.io/regenerator/)):

```
TypeError: unknown: Property right of AssignmentExpression expected node to be of a type ["Expression"] but instead got undefined
    at Object.validate (bundle.js:72816:11)
    at validateField (bundle.js:79750:9)
    at validate (bundle.js:79743:3)
    at validateNode (bundle.js:68534:27)
    at Object.assignmentExpression (bundle.js:64971:36)
    at Ep.assign (bundle.js:120582:34)
    at Ep.emitAssign (bundle.js:120575:18)
    at Ep.explodeViaTempVar (bundle.js:121138:19)
    at bundle.js:121250:23
    at Array.map (<anonymous>)
```

I'd like to make a PR to fix this problem later.

See also: babel/babel#15142

SuperSodaSea:
Fix crash when compiling `yield` or `await` in sparse array, and add test `should work when yield is in a sparse array` to verify the fix.
Fixes #627.

### DIFF
--- a/packages/transform/src/emit.js
+++ b/packages/transform/src/emit.js
@@ -1100,7 +1100,9 @@ Ep.explodeExpression = function(path, ignoreResult) {
   case "ArrayExpression":
     return finish(t.arrayExpression(
       path.get("elements").map(function(elemPath) {
-        if (elemPath.isSpreadElement()) {
+        if (!elemPath.node) {
+          return null;
+        } if (elemPath.isSpreadElement()) {
           return t.spreadElement(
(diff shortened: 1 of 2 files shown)


benjamn (approved):
Looks great, thanks @SuperSodaSea!

PR #630, opened 2022-11-10
Require `.next` method on delegate iterator

nicolo-ribaudo:
Fixes https://github.com/babel/babel/issues/15177


benjamn (approved):
Looks good, thank you @nicolo-ribaudo!

PR #636, opened 2022-11-24
Fix: `yield*` should throw if not iterable

### LINKED ISSUE #202: Sandbox bug report - "yield* undefined" doesn't throw (or is caught inside and hidden)
The following code does not transform as expected:

``` js
function* gen() {
  yield* undefined;
}

let y = gen();
while(true) {
  let caught = false;
  let r;
  try {
    r = y.next();
  }
  catch(e) {
    caught = true;
  }
  alert(caught ? "caught" : "not caught");
  if(r.done)
    break;
}
```

SuperSodaSea:
`yield*` an non-iterable value should throw `TypeError`, rather than treating it as an iterator with no values.

Fixes #202.
See also: babel/babel#1648, babel/babel#15172

### DIFF
--- a/packages/runtime/runtime.js
+++ b/packages/runtime/runtime.js
@@ -487,7 +487,7 @@ var runtime = (function (exports) {
   };
 
   function values(iterable) {
-    if (iterable) {
+    if (iterable || iterable === "") {
       var iteratorMethod = iterable[iteratorSymbol];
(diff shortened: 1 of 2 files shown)


benjamn (approved):
Thanks for this fix + tests, and sorry for the long wait @SuperSodaSea!

PR #680, opened 2023-07-20
Do not use `-1` to mark uninitialized location numeric literals

nicolo-ribaudo:
In Babel 8 we are going to improve validation in the `t.numericLiteral` builder to only allow valid numeric literals, and it will start throwing for `-1` (since `-1` should be represented as an unary expression).


benjamn (approved):
Thanks @nicolo-ribaudo!

Issue #703, opened 2023-10-02
Generator spec reference link is outdated in `runtime.js` file


rbals0445:
Hello! (related : https://github.com/facebook/regenerator/issues/158)

While looking at the regenerator code, I found that the link was not working properly.

https://github.com/facebook/regenerator/blob/main/packages/runtime/runtime.js#L263
![image](https://github.com/facebook/regenerator/assets/60285506/435b99e8-f326-417d-a17f-b210fc60a391)

I think you meant number three of GeneratorResume. so it would be nice to replace the link with an official document

> https://262.ecma-international.org/6.0/#sec-generatorresume

![image](https://github.com/facebook/regenerator/assets/60285506/2589ee34-4b9c-426d-869f-44d28ca0e5a8)

thank you!

PR #704, opened 2023-10-02
Fix outdated reference link in `runtime.js`

rbals0445:
related : https://github.com/facebook/regenerator/issues/703

reference link now has 404 error. so i change the link to validated spec document

thank you!!


benjamn (approved):
Thanks for catching this @rbals0445!

PR #711, opened 2023-12-04
Fix yield * of falsy values

nicolo-ribaudo:
Test case found while running test262 with Babel: https://github.com/tc39/test262/blob/467a0fe68f633fabde77df94d993c45e840627a0/test/language/expressions/yield/star-in-rltn-expr.js


benjamn (approved):
Makes sense to me, thanks @nicolo-ribaudo!

